Issue #9, opened 2019-02-01
Deployment tasks


reginafcompton:
We can handle some deployment tasks early on, mainly, spinning up a Just Spaces server (if we want to do this) and deploying a staging site. 

Before starting this issue, I'll touch base with the DevOps team re: https://github.com/datamade/devops/issues/66

reginafcompton:
We will deploy the Just Spaces staging site on the Staging server [(given our conversation during devops).](https://github.com/datamade/devops/issues/66#issuecomment-460327717)

@beamalsky - we're trying out [a new pattern for creating deployment scripts](https://github.com/datamade/deploy-a-site/pull/45): you'll be taking the freshly written documentation for a test drive! Follow the deploy-a-site documentation, as it appears in the [cookie-cutter branch](https://github.com/datamade/deploy-a-site/tree/cookiecutter). @hancush will look forward to any feedback you have!

beamalsky:
Exciting !!

Issue #8, opened 2019-02-01
Create user login


reginafcompton:
We can safely complete the first part of 3.1. Most of this should be out-of-the-box Django.

```
3.1 Admin backend
DataMade will setup a basic administrative interface that will allow users to create accounts, login, and manage data. The basic components will include:

* Account creation
* Secure user login
* Password retrieval
```

reginafcompton:
This issue has one element of complexity: users should belong to a particular Agency (from the PLDP).

I can imagine a solution, wherein the users select a Agency from a dropdown (with results generated from our database). We could have UCD pre-populated as an agency, for this initial version.

We also need to think about when and how we populate our database with Agencies. Is there public form that just auto-generates these organizations? Or do all participating Agencies need to be approved by UCD?

beamalsky:
I propose that in our first iterations we assume that DataMade/UCD will need to add new agencies

reginafcompton:
I concur.

reginafcompton:
We'll open a different issue for password retrieval and email set-up.

beamalsky:
Closed by #33

Issue #7, opened 2019-02-01
Create a Django project


reginafcompton:
Initiate a Django project for Just Spaces!

beamalsky:
Completed!

PR #10, opened 2019-02-04
Create a Django project

beamalsky:
Completes #7, pending review from @reginafcompton !

Waiting on wireframe review from UCD to start on views and templates. After this is complete I can start #8 and #9.


.gitignore
```diff
@@ -1,104 +1,9 @@
...
-
-# mypy
-.mypy_cache/
+.DS_Store
+/keyrings/live/pubring.gpg~
+/keyrings/live/pubring.kbx~
+/keyrings/live/secring.gpg
+*.pyc
+/justspaces/local_settings.py
+/configs/local_settings.py.travis
```
reginafcompton: You can use a couple different files for travis settings. 

1. The `deploy-a-site` documentation suggests using [the `test_config`](https://github.com/datamade/deploy-a-site/blob/master/Setup-deployment-hook.md#setup-codedeploy-hook-for-github--travis-ci). 

2. I used the [`local_settings.py.example` for travis in CES](https://github.com/datamade/coordinated-entry-screening/blob/master/.travis.yml#L23) (and I prefer that practice now), since it's a good way to check that your example file stays up-to-date. 

Either way, you probably do not want a travis-specific settings file, so as to avoid having *another* settings file to keep track of.

.gitignore
```diff
@@ -1,104 +1,9 @@
...
+/configs/local_settings.py.staging
+/configs/local_settings.py.production
```
reginafcompton: The deploy-a-site templates put `py` at the end of the file name:
https://github.com/datamade/deploy-a-site/blob/a20345bdc85125cd69b4f7891a6b11bd6e406a72/templates/django/%7B%7Bcookiecutter.directory_name%7D%7D/scripts/after_install.sh#L36

Let's follow the documented practice.

keyrings/live/blackbox-admins.txt
```diff
@@ -0,0 +1,2 @@
+beamalsky@gmail.com
+reginafcompton@gmail.com
```
reginafcompton: Either Derek or Forest should be on every project. Let's add Derek.

tests/test_example.py
```diff
@@ -0,0 +1,4 @@
+import pytest
```
reginafcompton: Hooray! You read my mind.

.gitignore
beamalsky: Oh I like your example file method!

keyrings/live/blackbox-admins.txt
beamalsky: Added!

tests/test_example.py
beamalsky: 🎉

.gitignore
reginafcompton: great - can you remove `local_settings.py.travis.gpg`? (don't forget to deregister it with blackbox, first! and also delete the directive for it in the gitignore.)

PR #14, opened 2019-02-06
Deployment tasks

beamalsky:


beamalsky:
Closes #9

.travis.yml
```diff
@@ -0,0 +1,36 @@
...
+- pip install --upgrade pip
+- pip install -r requirements.txt
+addons:
+  postgresql: '9.6'
+  apt:
+    packages:
+    - postgresql-9.6-postgis-2.3
+before_script:
+- psql -c 'create database "just-spaces";' -U postgres
+- psql -U postgres -d travis -c "create extension postgis"
```
reginafcompton: I don't think we need lines 17 and 18 to run the migrations: you would need them if your tests do not setup the database for you, I think. (The travis docs suggests doing exactly what you've done, but I think that's if your tests need the Postgres environment: https://docs.travis-ci.com/user/database-setup/#using-postgresql-in-your-builds.)

Can we try it without lines 17 and 18?

justspaces/local_settings.example.py
```diff
@@ -13,7 +13,7 @@
     'default': {
         'ENGINE': 'django.contrib.gis.db.backends.postgis',
         'NAME': 'just-spaces',
-        'USER': '',
+        'USER': 'postgres',
```
reginafcompton: This might change depending on my comment above, i.e., you  might be able to leave it as a blank string.

.travis.yml
beamalsky: Yep let me give it a shot! The deploy-a-site docs say to include `psql -U postgres -d travis -c "create extension postgis"` but not `psql -c 'create database "just-spaces";' -U postgres`. I did that first, but Travis complained about not seeing a database.

beamalsky: Result: `psycopg2.OperationalError: FATAL:  database "just-spaces" does not exist`

PR #17, opened 2019-02-14
Survey tool: Create

beamalsky:
The start of survey form building functionality. Surveys will have to interact with agencies, locations, and studies, so I've sketched out the basic model forms for each of those.


justspacesapp/static/css/custom.css
```diff
@@ -0,0 +1,268 @@
...
+/* ---------------------------------------------------
+    SIDEBAR STYLE
+----------------------------------------------------- */
+
+.wrapper {
+    display: flex;
+    align-items: stretch;
+}
+
+#sidebar {
```
reginafcompton: It's typically best practice to use classes for CSS styles. (You can find myriad discussions about this: a cursory search turned up [this article](https://dev.to/claireparker/reasons-not-to-use-ids-in-css-4ni4).) 

I'll leave it up to you, ultimately, but let me know what you think.

[N.b., I see that you use [`#sidebar` in the JS.](https://github.com/datamade/just-spaces/pull/17/files#diff-a9ed6d5e5bde2db9db7ce31a5712bda4R71) That makes sense.]

justspacesapp/static/css/custom.css
```diff
@@ -0,0 +1,268 @@
...
+
+#sidebar.active {
+    min-width: 80px;
+    max-width: 80px;
+    text-align: center;
+}
+
+#sidebar.active .sidebar-header h3,
+#sidebar.active .CTAs {
+    display: none;
```
reginafcompton: Would it be best to use the more modular Bootstrap 4 utilities for these styles (`display: none`, `display:block`, `text-align`)? It's certainly more conventional to follow the Bootstrap paradigm, but I am not sure how many individual elements you'd have to handle.

https://getbootstrap.com/docs/4.0/utilities/display/#notation

justspacesapp/templates/base.html
```diff
@@ -0,0 +1,43 @@
...
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <!-- <link rel="icon" href="{% static 'images/favicon.jpg' %}"> -->
+
+    <!-- JS -->
+    <script
+    src="https://code.jquery.com/jquery-2.2.4.min.js"
```
reginafcompton: Let's host a copy of the jquery.js, rather than using the CDN (then, we need not worry about ever losing access to the CDN). You can run `curl https://code.jquery.com/jquery-2.2.4.min.js` to get it.

reginafcompton: Oh, maybe related (?), what's that integrity property do?

justspacesapp/templates/base.html
```diff
@@ -0,0 +1,43 @@
...
+
+    <title>Just Spaces</title>
+  </head>
+
+  <body>
+
+      {% include "sidebar.html" %}
+
+          <!-- Page Content  -->
+          <div id="content">
```
reginafcompton: Why is `content` an ID and not a class?

justspacesapp/templates/create_study.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'base.html' %}
+{% load static %}
+
+{% block content %}
+  <h2>New Study</h2>
+  <form method="POST" class="post-form">{% csrf_token %}
+     {{ form.as_p }}
```
reginafcompton: For now, this should work just fine. But, eventually, we should think about the question of [browser-side form validation](https://pageclip.co/blog/2018-02-20-you-should-use-html5-form-validation.html) (particularly given the imminent complexity of our forms). We may need to add `use_required_attribute = False` to the form classes, if we want to prevent browser-side validation, but let's make that determination later. Just raising it here!

justspacesapp/forms.py
```diff
@@ -0,0 +1,23 @@
+from django import forms
+
+from pldp.models import Agency, Location, Study, Survey
+
+class CreateAgencyForm(forms.ModelForm):
+    class Meta:
+        model = Agency
+        fields = ('__all__')
```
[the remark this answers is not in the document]
beamalsky: Good q! I figure these forms will change and was making these quickly, but I'll keep an eye on that.

justspacesapp/static/css/custom.css
beamalsky: This was a good read, thanks! I'll switch them over.

justspacesapp/templates/create_study.html
beamalsky: Definitely! Leaving this open as a future reminder.

justspacesapp/static/css/custom.css
beamalsky: These styles show and hide the "Just Spaces" name and logo based on whether or not the menu is collapsed (ie, has the `active` class). I'm not sure how to get at that through bootstrap's styles. Maybe we could pair on that?

beamalsky:
@reginafcompton this is ready for your review! I now have a good handle on how to create custom plugins with fobi, and I've made form elements for age and gender. However, I'm realizing that the PLDP widgets for most of these are actually multiple inputs and I'd like to talk through with you how much to restrict users to the protocol. As this stands, this is enough to implement both sample surveys that UCD wanted to start with.

fobi_custom/plugins/pldp/form_elements/fields/single/age/fobi_form_elements.py
```diff
@@ -0,0 +1,36 @@
...
+
+    uid = "pldp_age_single"
+    name = "Age (single)"
+    form = PLDPAgeSingleForm
+    group = "Public Life Data Protocol"  # Group to which the plugin belongs to
+
+    def get_form_field_instances(self, request=None, form_entry=None,
+                                 form_element_entries=None, **kwargs):
+
+        choice_level = 'AGE_{}_CHOICES'.format(self.data.detail_level.upper())
```
reginafcompton: Nice way to do this!

fobi_custom/plugins/pldp/form_elements/fields/single/age/forms.py
```diff
@@ -0,0 +1,20 @@
...
+from fobi.base import BasePluginForm
+from pldp.forms import AGE_TYPE_CHOICES
+
+
+class PLDPAgeSingleForm(forms.Form, BasePluginForm):
+    """PLDPAgeSingleForm."""
+
+    plugin_data_fields = [
+        ("label", ""),
+        ("name", ""),
```
reginafcompton: I wonder if label and name should be pre-populated. It seems like the pldp plugins should be consistent, across all fields, not just the age type options. 

As for the age types, it's not entirely clear what I'm selecting. 

screen shot 2019-02-26 at 1 24 48 pm

We can't assume that our users have an intimate (or any knowledge) of the pldp. Is there a better way to communicate what the detail level means?

***

For the question of age, single vs. multiple ("how old are you?" vs. "how many people in this space fit into this age type?")....yes, as you said, I think we need two different form plugins. It seems like pldp was designed with the multiple input in mind. You can see how each type is a row, and the value of that row is an integer (i.e., the number of people). 

screen shot 2019-02-26 at 1 30 10 pm

I think we should build out both of these options before sharing with UCD, if you think that's possible within our given timeline. What do you think?

justspaces/urls.py
```diff
@@ -22,4 +22,11 @@
     url(r'^admin/', admin.site.urls),
     url(r'^pong/$', pong,),
     url(r'', include('justspacesapp.urls')),
+    # DB Store plugin URLs
+    url(r'^fobi/plugins/form-handlers/db-store/',
+        include('fobi.contrib.plugins.form_handlers.db_store.urls')),
+    # View URLs
+    url(r'^fobi/', include('fobi.urls.view')),
+    # Edit URLs
+    url(r'^fobi/', include('fobi.urls.edit')),
```
reginafcompton: Can we customize the template for editing a form? I do not think we need to show handlers.

screen shot 2019-02-26 at 1 37 51 pm

beamalsky: Agreed! I was going to leave this until I figure out how to set a default handler, because right now you have to add them manually.

justspacesapp/urls.py
```diff
@@ -15,13 +15,15 @@
...
 
 urlpatterns = [
-    url(r'^$', view=create_form_entry, name='home'),
+    url(r'^$', RunSurvey.as_view(), name='home'),
     url(r'create-agency$', CreateAgency.as_view(), name='create-agency'),
     url(r'create-location$', CreateLocation.as_view(), name='create-location'),
     url(r'create-study$', CreateStudy.as_view(), name='create-study'),
     url(r'create-survey$', view=create_form_entry, name='fobi.create_form_entry'),
+    url(r'edit-survey$', EditSurvey.as_view(), name='edit-survey'),
+    url(r'run-survey$', RunSurvey.as_view(), name='run-survey'),
```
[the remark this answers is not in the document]
beamalsky: Would you be able to sketch out what these URLs might look like?

justspacesapp/views.py
```diff
@@ -24,13 +25,23 @@ class CreateStudy(CreateView):
...
-    # to create a sample area:
-    # INSERT INTO pldp_studyarea (name, area) VALUES ('Example area', ST_SetSRID(ST_MakePoint(1,2),4326));
-
-    template_name = "create_survey.html"
-
-    def get_context_data(self, **kwargs):
-        context = super().get_context_data(**kwargs)
-        context['form'] = CreateSurveyForm()
-        return context
+# class CreateSurvey(TemplateView):
```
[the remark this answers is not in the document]
beamalsky: The Create Form view is from fobi, including the "name" and "is_public" fields. Could you say more about how you're envisioning this relationship being documented? Ie, how what works?

I've made modifications to the Fobi Create Form Form to take out fields, and I think these can be further modified to bring it in line with the needs of the repo: https://github.com/datamade/django-fobi/pull/1

justspacesapp/templates/edit_survey.html
```diff
@@ -0,0 +1,34 @@
...
+      <tr>
+        <th scope="col">Survey Name</th>
+        <th scope="col">Created</th>
+        <th scope="col">Updated</th>
+      </tr>
+    </thead>
+    <tbody>
+      {% for survey in surveys %}
+        <tr>
+          <td><b><a href="/fobi/forms/edit/{{survey.id}}">{{survey.name}}</a></b></td>
```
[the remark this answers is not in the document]
beamalsky: Yes! I agree and was planning to do this—I'm not sure why Fobi uses these labels as identifiers instead of something like a UUID, and may switch it over.

fobi_custom/plugins/pldp/form_elements/fields/single/age/forms.py
beamalsky: I can add help text to explain what the different detail levels mean, good call. Fobi isn't set up to handle form elements with multiple inputs. I'll start digging into that to see what it would take to implement.

justspacesapp/static/css/custom.css
reginafcompton: (Excuse my delayed response!) I think it's fine as it, in that case. No need to fuss with this.

justspacesapp/views.py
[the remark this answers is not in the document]
reginafcompton: Sure thing. We gain a lot from Fobi, and it would be useful to future maintainers of this repo (and me!) to explain how we leverage that tool. 

I think the initial pass of such an explanation should be in the README of this PR.

I could see a header in the README with something like, "Fobi views," and then explain which views Fobi gives us and link to relevant documentation. It can be concise! The aim: making it easy for newcomers to navigate your implementation of Fobi. 

Does that seem reasonable?

***
In addition, could you explain (for my own education!) what `is_public` means in Fobi? and how that relates to Just Spaces?

justspacesapp/urls.py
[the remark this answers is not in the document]
reginafcompton: Will do! Let's plan to chat about this tomorrow.

justspacesapp/templates/edit_survey.html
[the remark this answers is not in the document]
reginafcompton: Great. Do you think you could do that in this PR?

fobi_custom/plugins/pldp/form_elements/fields/single/age/forms.py
reginafcompton: "Fobi isn't set up to handle form elements with multiple inputs." Oh, shoot. You mentioned that. 

Is this a bad idea? We could have a form plugin for each age question (e.g., "How many people are 0-14"), and then group them together somehow....

justspaces/urls.py
reginafcompton: okay - got it.

justspacesapp/views.py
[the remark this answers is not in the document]
beamalsky: On it, I'll create a section in the README! I think that will be helpful for me to have to articulate as well. 

Currently, `is_public` is just a saved Boolean attached to a survey and isn't used anywhere in the app. It's built into Fobi as a value on the [FormEntry model](https://github.com/datamade/django-fobi/blob/just-spaces-edits/src/fobi/models.py#L364) (that is, the object you create for a new form). I haven't hidden it because I thought we might use it, but I've hidden others and could do the same. There's still work to be done to solidify the starting values associated with a survey—maybe I can prioritize that tomorrow.

justspacesapp/templates/edit_survey.html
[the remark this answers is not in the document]
beamalsky: Yep!

reginafcompton:
I am consolidating our various threads into a checklist! The following should be accomplished before merging this PR:

- [ ] "solidify the starting values associated with a survey": https://github.com/datamade/just-spaces/pull/17#discussion_r260518537 (n.b. I think we should remove `is_public` for now)
- [ ] initial documentation of how-we-use-Fobi in the README: https://github.com/datamade/just-spaces/pull/17#discussion_r260497820
- [ ] Override Fobi survey edit template and/or use UUID as an identifier: https://github.com/datamade/just-spaces/pull/17#discussion_r260399222
- [ ] Clarify "detail levels" with explanatory text: https://github.com/datamade/just-spaces/pull/17#discussion_r260427570

### Items to think about, but to implement in different PRs
- [ ] Map out RESTFUL urls: https://github.com/datamade/just-spaces/pull/17#discussion_r260397618 (I am assisting with this one)
- [ ] Identify a solution for handling "multiples" or fields with a value + count (e.g., "how many people in this space are 0-14?") – we can implement this solution in another PR: https://github.com/datamade/just-spaces/pull/17#discussion_r260427570

beamalsky:
To consider also doing in this PR: 
Form element plugins (eg, select multiple, radio, bool) are somewhat inconsistently implemented in fobi. Since I'm already going through them and modifying the `name` field to turn it into an automatically generated ID, it makes sense to me to also do a consistency pass on each element we're using to make sure it provides a field for help text and no extraneous fields.

beamalsky:
Fields from the [PLDP survey model](https://github.com/datamade/django-pldp/blob/master/pldp/models/survey.py#L9-L81) that could be part of a survey's "properties", as defined by fobi:

- location
- study

justspacesapp/urls.py
```diff
@@ -13,9 +13,98 @@
...
+    url(r'studies/create$',
+        StudyCreate.as_view(),
+        name='studies-create'),
+
+    # Create new survey
+    url(r'surveys/create$',
+        view=fobi.views.create_form_entry,
+        name='fobi.create_form_entry'),
+
+    # Create survey element
```
reginafcompton: I think this URL should be restructured. You have a survey, and the survey has elements that belong to it, right? If so, you're URL should be:

`surveys/{survey id}/elements/{element id}/create`

Otherwise, you have two nondescript IDs side-by-side. 

The same goes for `surveys/handlers` below.

justspacesapp/urls.py
```diff
@@ -13,9 +13,98 @@
...
+        view=fobi.views.edit_form_handler_entry,
+        name='fobi.edit_form_handler_entry'),
+
+    # Delete survey handler
+    url(_(r'^surveys/handlers/delete/(?P<form_handler_entry_id>\d+)/$'),
+        view=fobi.views.delete_form_handler_entry,
+        name='fobi.delete_form_handler_entry'),
+
+    # Edit survey
+    url(_(r'^surveys/edit/(?P<form_entry_id>\d+)/$'),
```
reginafcompton: More typically, the action comes after the id. (I think Fobi does it the other way around....and I may have done the same in my original PR.) 

But even more typically, the action does not appear in the URL, but rather in the request....[the first comment in this StackOverflow article nicely summarizes the difference.](https://stackoverflow.com/questions/256349/what-are-the-best-common-restful-url-verbs-and-actions) It looks like Fobi includes the action verb in their (function-based?) views, which might make standardizing these URLs a little tricky. 

Can you add a doc string flagging this? (We might want to revisit, as the tool develops.)

justspacesapp/urls.py
```diff
@@ -13,9 +13,98 @@
...
+        view=fobi.views.edit_form_entry,
+        name='fobi.edit_form_entry'),
+
+    # Delete survey
+    url(_(r'^surveys/delete/(?P<form_entry_id>\d+)/$'),
+        view=fobi.views.delete_form_entry,
+        name='delete_form_entry'),
+
+    # Survey list edit
```
reginafcompton: We want to use query params to distinguish between edit and run functionality.  That's the main point of this issue: https://github.com/datamade/just-spaces/issues/26

justspacesapp/urls.py
```diff
@@ -13,9 +13,98 @@
...
+
+    # Submitted survey generic
+    url(_(r'^surveys/view/submitted/$'),
+        view=fobi.views.form_entry_submitted,
+        name='fobi.form_entry_submitted'),
+
+    # Submitted survey detail
+    url(_(r'^surveys/view/(?P<form_entry_slug>[\w_\-]+)/submitted/$'),
+        view=fobi.views.form_entry_submitted,
+        name='fobi.form_entry_submitted'),
```
reginafcompton: Overall, I think we have a good start on RESTful routes. I noted a couple places that require revision, but please plan to revise in another PR. 

For now, please add the requested doc string (cf. above).

justspacesapp/urls.py
beamalsky: The elements themselves have a survey ID attribute, so it's not in the url. This is `surveys/elements/create/{element} id}/{type of element}`. For example: `/surveys/elements/create/23/boolean/`. Would you still suggest changing this?

justspacesapp/urls.py
beamalsky: Fobi does include actions in their function-based views and urls, and these are ported pretty directly. I'll leave a comment for now!

justspacesapp/urls.py
beamalsky: Ah, I see. Combining into one list for now since we don't have published or unpublished states attached to surveys yet.

justspacesapp/urls.py
reginafcompton: Oh! I see. To be clear: one element can belong to multiple surveys? and one survey can have multiple elements? i.e., an element does not belong to a particular survey...

If so, then we do not need `surveys` at the front of the URL. The url should be: `/elements/create/{element} id}/{type of element}`

justspacesapp/urls.py
reginafcompton: Yes, that's a pitfall of function-based views. We can talk about if it's worth fighting against.

justspacesapp/views.py
```diff
@@ -26,13 +26,7 @@ class StudyCreate(CreateView):
     success_url = '/'
 
 
-class SurveyListEdit(ListView):
+class SurveyList(ListView):
```
reginafcompton: This looks great! I'll open an issue to remind us to account for query params, i.e., to filter between published and unpublished (or whatever adjective we use).

beamalsky: Thank you!

justspacesapp/urls.py
beamalsky: An element belongs to just one survey, but surveys can have unlimited elements. Your example URL looks good and I'll switch it!

fobi_custom/plugins/pldp/form_elements/fields/age_multiple/forms.py
```diff
@@ -0,0 +1,61 @@
...
+                   ('complex', tuple(AGE_COMPLEX_CHOICES))))
+
+    label = forms.CharField(label="Label",
+                            required=True,
+                            help_text="Use this survey element to count the \
+                            number of people of a specific age range. Change \
+                            the question to match your selection from the \
+                            dropdown below.")
+    name = forms.CharField(required=True, widget=forms.widgets.HiddenInput())
+    age = forms.ChoiceField(choices=structured_choices,
```
reginafcompton: I think this provides a good initial solution! (And thank you for the explanatory help text.) We definitely have enough to consider in conversation with UCD, who will help us iron out the details (I'm sure).

Issue #28, opened 2019-03-01
Set up password retrieval via email for admin users


reginafcompton:
This issue will complete the remaining portions of 3.1 in the SOW.

beamalsky:
What do you think new users should be prohibited from doing until they verify their email?

reginafcompton:
My inclination would be to prohibit them from logging into the site...

But, on the other hand, perhaps we should allow users to signup *without an email*? Maybe we should check with UCD about user data expectations – will they need email addresses for admin users for any reason?

beamalsky:
What's the reason to allow users to skip entering an email? Users wouldn't have a way to reset their own passwords, and for what it's worth the PLDP requires an email attached to new studies.

reginafcompton:
A couple thoughts: (1) I think we can safely assume that all users will have an email (is that right?), and (2) I suppose we should not collect email addresses, unless we have a need for them.

With that said, the two reasons you give above strongly argue in favor of requiring an email. (It would be annoying if UCD or DataMade had to provide assistance with password resets, for users who did not provide email addresses.)

Further thoughts?

beamalsky:
I think password reset functionality is a need, as it's part of the scope! For the PLDP agencies we don't need to have an auto-populated email field (and in any case it should be change-able when that form is filled out), but it would be nice.

reginafcompton:
All right! Email shall be a required field for signup. 

...back to the comment above: what do think about prohibiting users from logging in, until they verify their email? It's straightforward way to ensure they provided a valid email, at the very least.

beamalsky:
That sounds good to me! I've never implemented that in Django but it seems like a pretty standard use case.

PR #33, opened 2019-03-04
Create user login and signup

beamalsky:
Addresses #33:

- Creates user login and logout pages
- Creates user creation view (`accounts/signup`) with Agency dropdown
- Sets permissions that user must be logged in to create, edit, or run, surveys
- For now, sets default logged in users as superusers. This will be changed in a separate PR once we talk to UCD about who should have permission to do what.


justspacesapp/views.py
```diff
@@ -30,3 +35,23 @@ class SurveyList(ListView):
     model = FormEntry
     template_name = "survey_list.html"
     context_object_name = 'surveys'
+
+class Signup(FormView):
```
reginafcompton: Do we need to expose a URL for Signup in the sidebar?

justspacesapp/views.py
```diff
@@ -30,3 +35,23 @@ class SurveyList(ListView):
...
+    template_name = "registration/signup.html"
+    form_class = JustSpacesUserCreationForm
+    success_url = '/'
+
+    def post(self, request):
+        superuser = JustSpacesUser(is_superuser=True)
+        form = self.form_class(request.POST, instance=superuser)
+        print("Here is the form:")
```
reginafcompton: Can we strike these print statements?

justspacesapp/urls.py
```diff
@@ -5,32 +5,41 @@
...
+from django.contrib.auth.decorators import login_required, permission_required
 from justspacesapp.views import AgencyCreate, LocationCreate, StudyCreate, \
-                                SurveyList
+                                SurveyList, Signup
 import fobi.views
 
 urlpatterns = [
     url(r'^$',
-        SurveyList.as_view(),
+        login_required(SurveyList.as_view()),
```
reginafcompton: It seems like, eventually, we'll want the home view to be a friendly public-facing page, e.g., something we build in Part 2. Can we name this URL "login"? Then, we can avoid having to edit this later on.

justspacesapp/views.py
```diff
@@ -30,3 +35,23 @@ class SurveyList(ListView):
...
+        print(form)
+        if form.is_valid():
+            form.save()
+            username = form.cleaned_data.get('username')
+            raw_password = form.cleaned_data.get('password1')
+            user = authenticate(username=username, password=raw_password)
+            login(request, user)
+            return redirect('home')
```
reginafcompton: I'd prefer that users do not get automatically logged into the site. Let's redirect to login.

Note! I also think we want users who sign-up to verify their accounts via email. I modified the relevant issue accordingly: https://github.com/datamade/just-spaces/issues/28

Do you agree?

justspacesapp/templates/registration/signup.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'base.html' %}
+
+{% block content %}
+  <h2>Sign up</h2>
+  <form method="post">
+    {% csrf_token %}
+    {{ form.as_p }}
```
reginafcompton: Currently, form errors do not get logged clearly: mainly, they appear in black, within a bulleted list. It would be best if the error message and corresponding form field appears in red.

You can do this several ways. (Though...we still need to decide if we want the browser to handle validating required fields.) Can you plan to implement a solution when you work on site design, this week?

justspacesapp/views.py
beamalsky: yes!

justspacesapp/views.py
beamalsky: also yes!

justspacesapp/views.py
beamalsky: Sure! I'll comment in the issue.

justspacesapp/templates/registration/signup.html
beamalsky: Definitely!

justspacesapp/urls.py
beamalsky: I'm intending this as a placeholder home page to use until we have the public-facing site views. I already have a url named login (`accounts/login`). I've changed the login view redirect to `survey-list` and added a comment to the home view. Does that address your concern here?

justspaces/settings.py
```diff
@@ -138,7 +138,7 @@
 STATIC_ROOT = os.path.join(BASE_DIR, 'static')
 STATIC_URL = '/static/'
 
-LOGIN_REDIRECT_URL = 'home'
+LOGIN_REDIRECT_URL = 'survey-list'
```
reginafcompton: I think this makes the most sense! (Later, we'll need to decide if this redirect should be to "editable" or "run-able" surveys.)

justspacesapp/urls.py
reginafcompton: yes!

Issue #55, opened 2019-03-13
Initiate app structure for public-facing analyses


reginafcompton:
@jeancochrane has [an R&D day scheduled for March 15](https://github.com/datamade/ops/issues/593). During this time she will:

* initiate an app for the public-facing dimension of Just Spaces: the app structure will a integrate a Babel, a JS compiler 
* refactor the less-precisely-named app `justspacesapp`

reginafcompton:
Ready to close! Thanks @jeancochrane and @beamalsky

Issue #29, opened 2019-03-04
Populate database with Clark Park and Porch survey examples


reginafcompton:
Create Clark Park and Porch surveys (using the current interface). 

This may require adding more form elements, so the surveys can reflect PLDP.

beamalsky:
Fields from the PLDP survey model that should eventually be implemented as form elements:

- [ ] location
- [ ] study
- [ ] time_start _use default time plugin to start_
- [ ] time_stop _use default time plugin to start_
- [ ] time_character _use default text plugin to start_
- [ ] representation _use default dropdown plugin to start_
- [ ] microclimate _use default text plugin to start_
- [ ] temperature_c _use default temp (?) plugin to start_
- [ ] method _use default dropdown plugin to start_

beamalsky:
Update: Clark Park and Porch surveys are built out on https://justspaces.datamade.us/ as examples! This should give us a good tool to talk through next steps with UCD. In particular, I'd like to talk to them about how closely they'd like to stick to the PLDP. I didn't end up needing to implement any of the above form elements for these sample surveys.

PR #57, opened 2019-03-15
Rename 'justspacesapp' to 'surveys'

jeancochrane:
## Overview

Rename the backend survey-collector app from `justspacesapp` to `surveys`.

### Notes

Unfortunately, doing this kind of app renaming can seriously mess with migrations. See: https://gist.github.com/jamesmfriedman/6168003

The way I got around this was simply by dropping and recreating the database following the instructions in the README. If we need to preserve data in the staging database, we can do something a little bit more sophisticated (dump out relevant data and reimport it), but it'd be nice to be able to skip that and save time if we can.

## Testing Instructions

 * Drop and recreate your database following the instructions in the README
 * Run `python manage.py test` and confirm that tests pass
 * Run `python manage.py runserver` and confirm that the app loads properly at localhost:8000
 * Run `grep -iRF "justspacesapp" --exclude-dir .git .` and confirm there are no references to the old app name in the code 

Connects #55


reginafcompton:
@jeancochrane - terrific! Thanks for making these adjustments.

For the database, the database on the staging site has some data, which we'd like to preserve. [See this issue.](https://github.com/datamade/just-spaces/issues/29)

It should not be too difficult to preserve select tables, since Bea prepared a data dump (versioned at the root of the `justspaces` project). Locally, I managed to restore the tables we need by doing the following.

1. In the original database, [find tables with data, which we need to dump and restore.](https://stackoverflow.com/questions/2596670/how-do-you-find-the-row-count-for-all-your-tables-in-postgres)
```
SELECT nspname AS schemaname,relname,reltuples
FROM pg_class C
LEFT JOIN pg_namespace N ON (N.oid = C.relnamespace)
WHERE
    nspname NOT IN ('pg_catalog', 'information_schema') AND
    relkind='r'
ORDER BY reltuples DESC;
```

```
 public     | spatial_ref_sys                               |      5435
 public     | countries_plus_country                        |       252
 public     | languages_plus_language                       |       184
 public     | auth_permission                               |       148
 public     | fobi_formelemententry                         |        44
 public     | django_migrations                             |        44
 public     | django_content_type                           |        37
 public     | fobi_formentry                                |         2
 public     | django_session                                |         2
 public     | justspacesapp_justspacesuser                  |         2
 public     | pldp_agency                                   |         1
...
```

2. Drop the old database, and make a new one. Run the new migrations.

3. Create a superuser.

4. Restore the tables we need from the versioned data dump. [(See table option)](https://www.postgresql.org/docs/current/app-pgdump.html)

```
pg_restore --data-only -d just-spaces -pldp_agency -t spatial_ref_sys -t countries_plus_country -t languages_plus_language -t fobi_formentry just-spaces.dump

# We can restore this table only after restoring `fobi_formentry`
pg_restore --data-only -d just-spaces -fobi_formelemententry just-spaces.dump
```

Jean, I'd love your thoughts on this! I also think we should touch base with @beamalsky as well.

beamalsky:
This method looks good to me, thank you for documenting @reginafcompton! Creating the two sample surveys on the current site took me about 45 min, so if this restoration gets gnarly for some reason it may be faster to just drop and recreate.

jeancochrane:
Alright, this has been deployed! I went ahead and did the database surgery myself @beamalsky -- let me know if anything looks off to you.

beamalsky:
Looks great, thank you @jeancochrane !

PR #59, opened 2019-03-15
Initialize frontend app with Parcel and ES6

jeancochrane:
## Overview

Initialize the Just Spaces frontend as a separate Django app. In the frontend app, scaffold a JavaScript dev environment with support for the following features:

- ES6 syntax with Babel
- Module bundling with Parcel
- Linting with JSHint
- Tests with Jasmine

### Notes

There's a lot going on here, so I'll leave comments on the diff.

## Testing Instructions

- Make sure you have Docker installed ([MacOS link](https://docs.docker.com/docker-for-mac/install/))
- Install dev Python dependencies with `pip install -U -r dev-requirements.txt`
- Run `docker-compose build` to build the NPM image
- In one shell, run `docker-compose up` to bundle and watch the frontend assets
- In another shell, run `python manage.py runserver` to run the dev server
- Navigate to http://localhost:8000, open the console, and confirm that you see the "Hello, World!" message
- Run Jasmine tests with `docker-compose run --rm frontend test`
- Run JSHint linting with `docker-compose run --rm frontend lint`

Connects #55, https://github.com/datamade/ops/issues/593


docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+  frontend:
+    build: ./frontend
+    volumes:
+      - ./frontend:/usr/local/src
+      - node_modules:/usr/local/src/node_modules
+    working_dir: /usr/local/src
+    entrypoint: yarn
+    command: start
+volumes:
+  node_modules:
```
jeancochrane: The named module here helps prevent `node_modules` from getting overwritten by the host OS when we mount `./frontend:/usr/local/src`.

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
+services:
+  frontend:
+    build: ./frontend
+    volumes:
+      - ./frontend:/usr/local/src
+      - node_modules:/usr/local/src/node_modules
+    working_dir: /usr/local/src
+    entrypoint: yarn
+    command: start
```
jeancochrane: In case you haven't used Node.js before, `yarn` is equivalent to `npm`. In addition, when you call `yarn` or `npm` with an argument that the CLI doesn't recognize (in this case, `start`), it will read the `scripts` section of the `package.json` and call the commands defined there.

frontend/.jshintrc
```diff
@@ -0,0 +1,44 @@
+{
+  // Set syntax to ES5 (e.g., complain about use of ES6 features)
+  "esversion": 9,
```
jeancochrane: This is the only modification I made to the recommended JSHint file here: https://github.com/datamade/node-style-guide/blob/master/.jshintrc This lets us lint the latest JS features.

frontend/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "devDependencies": {
+    "jasmine": "^3.3.1",
+    "jshint": "^2.10.2",
+    "parcel-bundler": "^1.12.2"
+  },
+  "scripts": {
+    "start": "parcel watch src/js/index.js --out-dir static/frontend/js --public-url /static/frontend/js --no-hmr",
+    "test": "parcel build tests/frontend/*.js --out-dir tests/compiled --no-source-maps && jasmine tests/compiled/*.js",
+    "lint": "jshint src/js/*.js tests/frontend/*.js",
+    "build": "parcel build src/js/index.js --out-dir static/frontend/js --public-url /static/frontend/js"
```
jeancochrane: Here are the custom scripts I've defined so that we can run e.g. `yarn build` (or more accurately, `docker-compose run --rm frontend build`) and get a production build without having to specify a super long command.

frontend/tests/frontend/helloworld.js
```diff
@@ -0,0 +1,9 @@
+import { helloWorld } from "../../src/js/helloworld";
+
+/* jshint undef: false */
```
jeancochrane: Without this comment, JSHint will get mad that we're using the global variables `describe`, `it`, etc. that are required by Jasmine.

frontend/tests/frontend/helloworld.js
```diff
@@ -0,0 +1,9 @@
+import { helloWorld } from "../../src/js/helloworld";
```
jeancochrane: We can only use ES6 imports in this Jasmine script because the `yarn test` command contains a compilation command that uses Parcel to compile this test file (see above). Otherwise, Jasmine won't understand ES6 syntax.

scripts/after_install.sh
```diff
@@ -30,7 +30,11 @@ sudo -H -u datamade $VENV_DIR/bin/pip install --upgrade setuptools
 
 # Install the project requirements into the deployment specific virtual
 # environment.
-sudo -H -u datamade $VENV_DIR/bin/pip install -r $PROJECT_DIR/requirements.txt --upgrade
+sudo -H -u datamade $VENV_DIR/bin/pip install -r $PROJECT_DIR/requirements.txt -r $PROJECT_DIR/dev-requirements.txt --upgrade
+
+# Install NPM dependencies.
+$VENV_DIR/bin/docker-compose build
+$VENV_DIR/bin/docker-compose run --rm frontend build
```
jeancochrane: Note that we'll have to make sure Docker is installed on the servers before this will run properly.

.gitignore
```diff
@@ -6,3 +6,12 @@
 /justspaces/local_settings.py
 /configs/local_settings.production.py
 /configs/local_settings.staging.py
+
+# JavaScript bundling
+.cache
+frontend/static/frontend/js/*
```
jeancochrane: The idea here is that we'll be compiling JS from the `frontend/src/js` directory to the `frontend/static/frontend/js` directory, the latter of which will allow Django to understand that it's a static file and serve it properly. We want to keep the compiled static files out of version control, because they're fully derived from the source files -- the source files should be the single point of truth, and all consumers of the code (dev, CI, staging/production) should compile those source files instead of retrieving compiled files from version control.

README.md
```diff
@@ -19,10 +20,10 @@
     mkvirtualenv -p python3 just-spaces
     ```
 
-3. Install the requirements.
+3. Install the Django requirements.
 
     ```bash
-    pip install -r requirements.txt
+    pip install -r requirements.txt -r dev-requirements.txt
```
reginafcompton: Could you tell me a bit about the advantages of having, er, "topical" `requirements.txt` files?

README.md
```diff
@@ -31,29 +32,39 @@
     cp just-spaces/local_settings.example.py just-spaces/local_settings.py
     ```
 
-5. Create the database:
+5. Install NPM dependencies for the frontend:
```
reginafcompton: I'd love to have more robust documentation of what's going on here! `just-spaces` will be the first implementation of this approach, and I imagine the project repo will be a valuable model for others. 

I think you could include that documentation (e.g., links to useful resources, verbose explanations) in the step-by-step, or you could have a subsection that gives an overview.

docker-compose.yml
reginafcompton: What do you think about adding a brief comment that explains the `start` command? Mainly, a quick way to direct someone from `docker-compose.yml` to `frontend/package.json`.

docker-compose.yml
reginafcompton: For my own knowledge! Could you say more about why `node_modules` would be overwritten, otherwise? Or perhaps point me to a useful resource?

frontend/.jshintrc
reginafcompton: Great!

reginafcompton: It will likely be necessary to update our style guide, if we decide to use this pattern company-wide. Maybe make a note of that here: https://github.com/datamade/ops/issues/593?

scripts/after_install.sh
reginafcompton: We plan to spin up a designated server for just-spaces production (when the time comes): it can certainly have docker. 

The staging site has `Docker version 18.09.1, build 4c52b90`.

justspacesapp/urls.py
```diff
@@ -13,11 +13,6 @@
 import fobi.views
 
 urlpatterns = [
-    # placeholder view for a public-facing landing page
```
reginafcompton: @beamalsky - are we okay removing this placeholder now? I know it served a purpose for the demo site for UCD, i.e., getting content on a home page. But it seems like we can do away with it?

frontend/package.json
reginafcompton: Again, to satisfy my curiosity: do you have a recommended resource for learning about this approach?

frontend/src/js/index.js
```diff
@@ -0,0 +1,3 @@
+import { helloWorld } from "./helloworld";
```
reginafcompton: Is it possible to import multiple functions within the curly brackets? if so, what would that look like? also: is it best practice to import individual functions, or could you import all functions in a js file, e.g., `import * from "./helloworld"`?

docker-compose.yml
jeancochrane: Sure! The gist of it is that `node_modules` is a directory that gets created automatically by `npm isntall`/`yarn install` to hold dependencies (kind of like `site-packages` in Python) and it needs to live in the application directory in order to be importable into the application code. There are two ways to manage it in a container:

1. Run `npm install` or `yarn install` in the Dockerfile as a build step, meaning that `node_modules` is pre-buillt in the application container image and doesn't exist on the host OS
2. Run `docker-compose run frontend install` after the container image has already been built, meaning that `node_modules` will get installed on the host OS and then passed into the container when the application directory is mounted as a volume

In my experience 1) is preferable to 2), since NPM packages [are sometimes different for different operating systems](https://stackoverflow.com/a/35238945), and the Docker host OS is often different from the guest OS. So if you install packages for the host OS and then pass them to the container, sometimes you wind up with the wrong package for the container OS.

But 1) comes with the snag that when you mount the application directory from the host OS to the container (which you need to do in order for your local changes to be visible in the container) it will overwrite that directory in the container. Since 1) installs the `node_modules` in the container and not on the host OS, mounting the directory from the host OS to the container will overwrite the dependencies.

The named volume gets around this by creating a "third" location on disk where the volume will be stored. This means we can safely mount `frontend` to `/usr/local/src/` in the container without worrying about `usr/local/src/node_modules` getting overwritten, because `/usr/local/src/node_modules` is actually linked to the third location where the named volume is stored.

For more on the different types of volumes, see: https://success.docker.com/article/different-types-of-volumes Unfortunately I haven't found good docs on this particular solution to the problem, but it's what I've cobbled together after asking a few friends that use Docker and NPM frequently!

reginafcompton: Wow, thanks! 

In sum...the Dockerfile builds `node_modules` into the container. But (oh no!) we still have to mount the application directory to the container (to makes changes present in the container itself): this essentially undoes the work of `RUN yarn install`. So, we sneakily store the `node_modules` elsewhere – a named volume (`node_modules`) linked to `/usr/local/src/node_modules`. Very clever.

justspacesapp/urls.py
reginafcompton: I think that's a good approach @jeancochrane. Could you do that on this PR? And when we're ready, we can remove the placeholder, once-and-for-all.

reginafcompton:
@jeancochrane thank you for the thoughtful, robust answers to my questions! To get this PR merge-worthy, I think we have three items:

1. Add some documentation of the overall approach in the README: https://github.com/datamade/just-spaces/pull/59#discussion_r266166152. I'd also love more inline comments! (For example, [placing this PR comment](https://github.com/datamade/just-spaces/pull/59#discussion_r266166668) in the JS file.) But I'll leave that to you...Do you prefer to keep inline comments light, and instead, preserve explanations in the PR (as we've done above)? 
2. Add a comment about `start`: https://github.com/datamade/just-spaces/pull/59#discussion_r266626655
3. Adjust the location of our placeholder: https://github.com/datamade/just-spaces/pull/59#discussion_r267041845

Does that sound right?

jeancochrane:
Thanks for the clear list of action items @reginafcompton! I pushed changes for all 3 points.

I tried to add a few more inline comments, too. To your question about commenting: In general I assume that a dev reading my comments will know how all the tools work, and I tend to use comments to document code that is:

1. Tricky or unusual
2. Difficult to follow (contains necessary complexity)
3. Does not take the obvious solution, but for a good reason

For this PR I left a lot of GitHub comments because I knew these tools would be unfamiliar to reviewers, but I didn't leave as many in-code comments because most of what I'm doing is standard practice for the stack. In practice I tend to be more liberal with GitHub comments.

Does that make sense? There are a lot of different competing philosophies of comments, this is just the one I've grown accustomed to.

README.md
jeancochrane: Hm, I thought I pushed this documentation up, but the last commit in this PR seems to have not included the changes. (Just changed editors, and I'm still getting used to its Git workflow.) I'll just push those docs straight to `master` if that's OK with you @reginafcompton, since it's just an edit to the README.

reginafcompton: Go right ahead! Thanks for the catch.

Issue #63, opened 2019-03-26
Handle PLDP Survey metadata 


beamalsky:
In its current implementation, the `django-pldp` [Survey element](https://github.com/datamade/django-pldp/blob/master/pldp/models/survey.py#L11-L68) has the following fields:

- location (required)
- study (required)
- time_start
- time_stop
- time_character
- representation
- microclimate
- temperature_c
- method

Some of these (study, location) should stay the same between administrations of the survey. Others (time_start, time_stop, time_character, microclimate, temperature_c) should change. I'm not sure which of those categories representation and method would best fit into. Let's make sure we're all on the same page about those categories.

I first outlined this issue here: https://github.com/datamade/just-spaces/pull/17#issuecomment-467909992

In #56 I've implemented study, representation, and method as Fobi form elements. This means that they're added to surveys like any other question, the survey creator sets a default value for them, and then the survey runner can change them if they'd like. 

I see two paths here: 

1. Study and location stay as form elements and I modify them so that they cannot be changed while running the survey. As these are required fields, don't allow users to save surveys that don't include these questions. 
2. Move study and location into the [new survey creation form](https://justspaces.datamade.us/surveys/create) and save them as metadata there, meaning modifying the `django-fobi` FormEntry model. This is a heavier lift as it requires modifying the package, but may fit our data better and be worth the extra time.

beamalsky:
A third path we're pursuing: subclassing `FormEntry` to add our own fields! Working on it in #83. This means that some PLDP elements can be attached to the survey model and some can be components.

On survey model:

- [ ] location
- [ ] study

Components:

- [ ] total - according to PLDP, this will need to be a required field on all observational surveys
- [ ] time_start
- [ ] time_stop - right now I'm using this to record submission time. If this component exists we could override the submission time, or they could be separate
- [ ] time_character
- [ ] representation
- [ ] microclimate
- [ ] temperature_c
- [ ] method

beamalsky:
@jeancochrane this issue is just for picking an approach, which I believe we've done. Did you want to include implementing this approach in this sprint as well? It's partially done but there's more to do (see checkboxes above)

jeancochrane:
My instinct is to go ahead and implement this, since subclassing `fobi.models.FormEntry` turned out to be easier than anticipated.

beamalsky:
Closed by #87

Issue #25, opened 2019-02-26
Survey tool: Custom Handler Plugin


beamalsky:
We'll need a way to connect data collected by the fobi surveys forms with the Just Spaces models, which (will) follow the structure implied by the PLDP.

I'll build a custom fobi form handler plugin handler: https://django-fobi.readthedocs.io/en/0.13.8/#creating-a-new-form-handler-plugin

beamalsky:
I spent part of yesterday getting acquainted with Fobi data handling and[ modifying the package's form creation function so that it automatically attaches a default data saving method to new forms](https://github.com/datamade/django-fobi/pull/1/commits/0e161712232c0afbc0cd3d9ca8d44193bdc4c96b). Now that I better understand the Fobi defaults, I'd like to zoom out, get my thoughts in order, and talk about app-wide data modeling before I proceed.

## Data Handling, two paths forward 

### Option 1: Use the Fobi method of storing data and translate it into the Just Spaces models later

The Fobi handler has one model for saving form data: `SavedFormDataEntry`. It has these fields:

- `id`
- `user`: the user who submitted the form
- `form_data_headers`: a stringified JSON object with each form element `{'form_element_name':'form_element_label'}`
- `saved_data`: a stringified JSON object with each form element `{'form_element_name':'saved_data'}`
- `created`: date and time form was submitted
- `form_entry`: foreign key of the form that was submitted

This captures the information we need if we assume that a form will not change once it's been submitted (which I think we are, but we'd have to be very careful!). It plays nicely with Fobi's built in export views. On the other hand, those JSON blobs are clunky and far from the data structure described by the PLDP, and I fear they would make Part 2 of this project harder than necessary.

### Option 2: Modify the Fobi data handler to store more information about elements and line up with the PLDP
The PLDP has separate models for Survey and SurveyComponent, which we could employ here. 

Survey (corresponds with Fobi's SavedFormDataEntry)
- `id`
- `user`
- `created`
- `form_entry`

SurveyComponent
- `id`
- `survey`: foreign key
- `component_name`
- `component_label`
- `component_uid`: the type of component, eg boolean or pldp_age
- `position`: refers to question order in the survey
- `saved_data`

This should be something that's usable directly for the analyses in part 2 of Just Spaces, and is a more normalized and less JSON-y structure. It means that surveys could be edited and contextual information about entered data won't be lost. It would likely also mean more work to use other parts of the Fobi package, as it expects data to be structured a certain way.

[...]

PR #64, opened 2019-03-27
Data handler

beamalsky:
## Overview

This PR implements a custom Fobi data handler to store collected survey data in PLDP-described models. 

### Notes

See comments in `fobi_form_handlers.py` for details on the parts of this that will need more work once we address #63. Study, Location, and Total fields are handled with static placeholder values for now.

Something to consider in `django-pldp`: now that SurveyComponent is more generic, it no longer makes sense to have a detail_level field that defaults to "basic". Maybe we should switch that to None?

Also, this needs some tests! @reginafcompton has kindly agreed to help me get started on that in the next few days, so I can pick it up on Monday.

## Testing Instructions

* Make sure to update requirements and pull in new migrations! Several things have changed in `django-pldp`
* Create a new survey locally. Check to make sure it has "Collect data" as a handler added automatically
Screen Shot 2019-03-27 at 4 58 36 PM

* Add at least one sample question to your new survey
* Run the survey! Submit an answer to your question
* Check the database. Do you see what you just submitted in the `pldp_survey`, `pldp_surveyrow`, and `pldp_surveycomponent` tables?

Connects #25


beamalsky:
Moving this out of draft mode so that the Travis builds will run 👀

reginafcompton:
For tests, I outlined a possible way forward in this PR: https://github.com/datamade/just-spaces/pull/68

I suggest that you write unit tests for the `run` function of `CollectDataPlugin`. I set-up the basic pattern for doing that, but I did not stub out actual tests. 

Where to begin? Before I write tests, I like to ask: what does this function do? In this case, the `run` function creates instances of SurveyRows and SurveyComponents. Then, I like to ask: what could go wrong? Why would the `run` function not behave as expected? Those sorts of questions can help you identify edge cases, improve your code, and write more robust tests. 

Let me know what you think. Happy to pair on this, too!

beamalsky:
Hello @reginafcompton! I'll respond to your other feedback and very helpful test setup PR in a moment, but first let's make sure you can get this running locally. I forgot to make note of how to create a sample Location object—my apologies. I've gone ahead and made a data dump with the updated migrations, data handler, and sample metadata for the PLDP (Sample Study Area, Sample Area, Sample Location, and Sample Agency), which we can use on the staging site next time we're pushing to master. This involved deleting some outdated survey elements: `pldp_age_single`, `pldp_age_multiple`, `pldp_gender_single`, `pldp_gender_multiple`. These will need to be added back once we've completed [issue about 2 survey paths] and recreated those elements. 

## New Testing Instructions

- Drop the database: `dropdb just-spaces`
- Recreate the database: `pg_restore -C -j4 --no-owner just-spaces.dump | psql`
- Run the server: `python manage.py runserver`
- Login is the same as the staging site: testuser / 4SPUDO6f9Zm
- You should see the two example surveys: Porch Survey and Movies in Clark Park. Choose one of them, fill out some answers, and submit it.
- Check the database. Do you see what you just submitted in the pldp_survey, pldp_surveyrow, and pldp_surveycomponent tables?

I've also created a fixture—`pldp_sample`—that can reload a Sample Study Area, Sample Area, Sample Location, and Sample Agency in case we want that handy in the future. You shouldn't need to load that now, but for future reference the command is `python manage.py loaddata pldp_sample.json`

beamalsky:
@reginafcompton I agree 100% that that tabbed view you screenshoted (Elements - Handlers - Properties - Service) should change, and not only because those terms won't really mean anything to users of the tool. I think Handlers and Service can definitely be hidden, but Properties depends on the outcome of #63 so I was waiting to do anything to that UI until we address our handling of Survey metadata.

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
```diff
@@ -0,0 +1,81 @@
...
+        new_survey = Survey.objects.create(
+            study=study,
+            form_id=form_id,
+            location=location,
+            time_stop=time_stop
+        )
+
+        print("Survey submitted " + str(time_stop))
+
+        new_survey_row = SurveyRow.objects.create(
```
reginafcompton: Should the creation of a SurveyRow be nested under the iteration over `form_elements` (lone 53 ff.)? If I understand correctly, we'll create a survey row (with the number people counted by the question) with each survey component? 

I also wonder if the `total` default for intercept surveys should be 1. Can you say more about your reasoning?

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
```diff
@@ -0,0 +1,81 @@
...
+        time_stop = datetime.datetime.now(tz=get_current_timezone())
+
+        new_survey = Survey.objects.create(
+            study=study,
+            form_id=form_id,
+            location=location,
+            time_stop=time_stop
+        )
+
+        print("Survey submitted " + str(time_stop))
```
reginafcompton: Let's discard this print statement before merging the PR.

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
beamalsky: Hm, excellent question! The separate SurveyRow and SurveyComponent models come from Eric's original implementation of `django-pldp`, and I've been following that: the assumption is that each submitted survey (not each question in it) will have a required field with the total number of people being counted in that survey. If it's an intercept survey, with just one respondent, the total number of people counted would be 1. This will be useful in later analysis, when we want to make claims like "x% of people were riding bicycles."

Now that we have more example PLDP documents in Gehl's repo, I can check those assumptions! It seems this understanding is supported in their suggested [data structure](https://github.com/gehl-institute/pldp/blob/master/Public%20Life%20Data%20Protocol/PLDP_1.0%20Data%20Sheet%20Template%20(Single%20Table)). 

However, not all of their sample data (like [this one](https://github.com/gehl-institute/pldp/blob/master/Sample%20Data/sf-actscan.csv)) includes that total field at all. Maybe the total default should be None, and it should be assigned to 1 for intercept surveys specifically?

beamalsky: > The expectation is that the total number of people counted in observational surveys does not change from question-to-question.

Yes!

A reason to save 1 as a `total` for intercept surveys is that this would make it possible to compare across multiple submissions of the same survey form. For example, noting that out of 50 surveys, 40 of them responded "yes" to "Are you riding a bicycle?"

This is a question of how we're handling analysis, and it could also be derived later from noting that the survey is intercept. However, it makes more sense to me to save that in the `total` field while we're submitting the form and make a distinction between a survey regarding a single person and a survey regarding an unknown number of people.

beamalsky:
@reginafcompton I have a working test here that creates a sample survey, surveyrow, and surveycomponent, then checks that they're all existent and linked. 

I struggled to actually populate the sample survey with submitted data. Here is what I want it to have: 

`form.cleaned_data['c75e27cf-4d8f-49dc-bb15-da8137dac247'] = 5`

tests/test_data_handling.py
```diff
@@ -0,0 +1,28 @@
...
+from fobi.models import FormEntry
+
+from pldp.models import Survey, SurveyRow, SurveyComponent
+
+@pytest.mark.django_db
+def test_data_handler(sample_form_entry, sample_location, sample_study, sample_form_element, mocker):
+    request = mocker.MagicMock(spec=requests.Response)
+    form = assemble_form_class(form_entry=sample_form_entry)
+
+    # form.cleaned_data['c75e27cf-4d8f-49dc-bb15-da8137dac247'] = 5
```
reginafcompton: All right. This is tricky. I looked at [`django-fobi` for inspiration.](https://github.com/barseghyanartur/django-fobi/blob/070ee3239cc4df3f5e841ee8649d8c26a10f007e/src/fobi/tests/test_dynamic_forms.py#L35)

```
# Instantiate an instance of the `assemble_form_class`.
form_class = assemble_form_class(form_entry=sample_form_entry)

# Pass data to it
form = form_class(data={'c75e27cf-4d8f-49dc-bb15-da8137dac247': 5})

assert form.is_valid() == True
```

I have a couple questions about this:
* I do not understand the GUIDs as form fields. Do you know what fobi has going on here?
* What is your motivation for creating a `DynamicForm` with data? Your `run` function does not appear to do anything with it directly, right?

tests/test_data_handling.py
```diff
@@ -0,0 +1,28 @@
...
+    # A valid CollectDataPlugin requires a DynamicForm, a FormEntry, and a post request
+    plugin = CollectDataPlugin()
+    plugin.run(form=form,
+               form_entry=sample_form_entry,
+               request=request)
+
+    # Check that a survey has been created with a linked row and component
+    survey = Survey.objects.first()
+    row = SurveyRow.objects.filter(survey=survey)
+    component = SurveyComponent.objects.filter(row=row[0])
```
reginafcompton: Can you save `row[0]` in a variable? That would help future maintainers (and future you) understand what you're filtering on.

tests/test_data_handling.py
```diff
@@ -0,0 +1,28 @@
...
+
+    assert (len(row), len(component)) == (1, 1)
```
reginafcompton: I think this is a good assertion to make. But I'd also like you to make some assertions about the content of the newly created entities. Does the SurveyRow and SurveyComponent have the data you expect, given your fixtures?

tests/conftest.py
```diff
@@ -0,0 +1,106 @@
...
+        call_command('initialize_pldp')
+
+
+@pytest.fixture
+@pytest.mark.django_db
+def sample_user(db):
+    sample_user = JustSpacesUser.objects.create(
+        username='sampleuser'
+    )
+    sample_user.save()
```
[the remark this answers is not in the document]
beamalsky: eep! thanks

tests/test_data_handling.py
beamalsky: - Fobi has a 'name' field on each form element, which it uses to connect saved submitted data with other form data (eg plugin_uid, position, etc.). It's not the way I would have set up the package, necessarily, but changing it would mean doing some pretty major surgery to Fobi's base functions. It's essentially an additional ID for each form element, as I understand it.
- The run function saves submitted data! It should end up in the `saved_data` column of SurveyComponent. See: https://github.com/datamade/just-spaces/blob/data-handler/fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py#L67

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
beamalsky: @reginafcompton what do you think about this?

tests/test_data_handling.py
beamalsky: yep!

tests/test_data_handling.py
beamalsky: I added a lot more! This now checks for all fields that should be saved. Is this too much? It seems like too much for a single test to handle but I'm not sure how to break it up.

tests/test_data_handling.py
```diff
@@ -0,0 +1,64 @@
...
+    # find its linked components
+    row = rows[0]
+    components = SurveyComponent.objects.filter(row=row)
+    component = components[0]
+
+    # Check that the handler has created a single row and linked component
+    assert len(rows) == 1
+    assert len(components) == 1
+
+    # Check survey has the correct values
```
reginafcompton: Most of this code does not need the sort of obvious commenting you have here. 

If you feel like the code needs some explanation, you should place that in a docstring at the top of the text, i.e., right after line 12. It's not a requirement of this PR. Just something to think about.

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
reginafcompton: Okay, I think 1 as default for intercept surveys makes sense, given your example above. Let's plan for that. And, as you note, this particular detail may change depending on how the tool handles analysis.

PR #69, opened 2019-04-02
Draft data import pipeline

### LINKED ISSUE #67: Create import pipeline for static ACS data
While we may eventually want to allow admin users to dynamically import ACS data for custom geographies, most of the data should be statically loaded into the database ahead of time. This includes data at the following resolutions:

- National
- State (PA)
- County (Philadelphia)
- Municipal (Philadelphia)
- Census tract (within Philly)
- Census block (within Philly)

Create an import pipeline for the data outlined in https://github.com/datamade/just-spaces/issues/51#issuecomment-473122287 using:

- GNU Make for extract/transform
- [Census API Python client](https://github.com/datamade/census) for retrieving ACS tables
- Django management command for loading the data into the database

jeancochrane:
## Overview

Draft a simple data import pipeline for extracting a variety of ACS data for the app.

### Notes

For now, I didn't put the raw data under version control, since we're not totally sure what intermediate/final data will look like yet. I figure it would make sense to do that once the full pipeline is in a semi-finished state.

## Testing Instructions

* Follow the instructions in the README to make the data: https://github.com/datamade/just-spaces/blob/jfc/data-import-pipeline%2367/data/README.md
* Do a quick sniff of the data to make sure that it looks reasonable

Closes #67


data/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+COPY requirements.txt /usr/local/src/
+
+RUN set -ex \
+	&& buildDeps=" \
+		build-essential \
+	" \
+	&& deps=" \
+		unzip \
+		make \
+		wget \
```
jeancochrane: We don't need these dependencies immediately, but we'll probably need them once we start collecting shapefiles.

data/scripts/import_acs_data.py
```diff
@@ -0,0 +1,110 @@
...
+        self.writer.writeheader()
+
+    def write_acs_row(self, fips, acs_row):
+        row = {'fips': fips}
+        for var_code, var_stat in acs_row.items():
+            # Skip fields that don't match queried variables
+            if var_code not in self.codes_to_names.keys():
+                continue
+            # Process nulls
+            if var_stat == -666666666:
```
jeancochrane: I couldn't find any docs on this in the Census API. I was getting a few responses that looked like this from 2016 data, so I went ahead and treated them like nulls. I've since switched to 2017 data and I'm not sure whether the unusual responses persisted.

data/scripts/import_acs_data.py
```diff
@@ -0,0 +1,110 @@
...
+    pa_res = c.acs5.state(codes, PA_FIPS)
+    assert len(pa_res) > 0
+    for row in pa_res:
+        writer.write_acs_row(PA_FIPS, row)
+
+    # County-level data
+    philly_res = c.acs5.state_county(codes, PA_FIPS, PHILLY_FIPS)
+    assert len(philly_res) > 0
+    for row in philly_res:
+        writer.write_acs_row(PA_FIPS + PHILLY_FIPS, row)
```
jeancochrane: I'm not totally confident what the convention is for notating sub-state FIPS codes. This code is written assuming that it's the concatenation of all the containing geometries.

data/scripts/import_acs_data.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Tract-level data
+    tract_res = c.acs5.state_county_tract(codes, PA_FIPS, PHILLY_FIPS, '*')
+    assert len(tract_res) > 0
+    for row in tract_res:
+        fips = PA_FIPS + PHILLY_FIPS + row['tract']
+        writer.write_acs_row(fips, row)
+
+    # Blockgroup-level data
+    block_res = c.acs5.state_county_blockgroup(codes, PA_FIPS, PHILLY_FIPS, '*')
```
jeancochrane: A few of these variables aren't available at blockgroup resolution for the 5-year estimates. Rather than make this distinction in the import code, I decided to go ahead and structure the data the same way (i.e. each variable CSV file has the same FIPS codes) on the assumption that the application code will handle checking for null rows and warning that the blockgroup resolution is unavailable.

data/variables.yml
```diff
@@ -0,0 +1,598 @@
+age_and_sex:
+  verbose_name: Age and Sex
```
jeancochrane: None of the verbose names in this file are currently being used, but I assume at some point we'll want to use them to create a crosswalk for user display.

jeancochrane:
Requested review from @fgregg who may be able to accurately assess my use of https://github.com/datamade/census.

data/scripts/import_acs_data.py
fgregg: this would be good to report upstream.

data/variables.yml
fgregg: what is yaml giving us that just having this data as python dict is not?

data/scripts/import_acs_data.py
```diff
@@ -0,0 +1,111 @@
...
+        writer.write_acs_row(US_FIPS, row)
+
+    # State-level data
+    pa_res = c.acs5.state(codes, PA_FIPS)
+    assert len(pa_res) > 0
+    for row in pa_res:
+        writer.write_acs_row(PA_FIPS, row)
+
+    # County-level data
+    philly_res = c.acs5.state_county(codes, PA_FIPS, PHILLY_FIPS)
```
fgregg: why is philadelphia county the right unit, and not the city of philadelphia?

If philadelphia county is the right unti, I would call this philly_county_fips.

data/scripts/import_acs_data.py
```diff
@@ -0,0 +1,111 @@
...
+    assert len(philly_res) > 0
```
fgregg: that you need this assertion suggests a problem upstream?

fgregg:
Nice looking code!

data/scripts/import_acs_data.py
jeancochrane: Good call, I'll open an issue.

data/scripts/import_acs_data.py
jeancochrane: Is there an ACS geography that corresponds to municipalities like Philly? I didn't immediately see one in the `census` docs. I went with the county because the county and city are coterminous: https://en.wikipedia.org/wiki/Philadelphia_County,_Pennsylvania

data/scripts/import_acs_data.py
jeancochrane: I didn't run into any problems with empty results here, I just wanted to be defensive and have it complain loudly in case of an obvious error. Does that make sense?

data/variables.yml
jeancochrane: Two primary considerations:

1. YAML is easier to write than Python dicts or JSON, and I had to write this out by hand
2. There's enough data here that it needs to be in its own separate file to avoid cluttering up the code, so we might as well store it in a language-agnostic data format

data/scripts/import_acs_data.py
jeancochrane: Opened https://github.com/datamade/census/issues/72 for this.

data/variables.yml
fgregg: okay

data/scripts/import_acs_data.py
fgregg: Named places. This was one of the reasons why we originally wrote https://github.com/datamade/census_area/#usage

Newer versions of the Census API actually have native support for named places, which we don't have convenience methods for in our census wrapper. (https://github.com/uscensusbureau/api-geoHierarchy-changes/blob/master/changes.md?eml=gd&utm_medium=email&utm_source=govdelivery)

Since they are coterminous, this seems fine. It would be good to add a comment about that fact.

PR #70, opened 2019-04-03
Blackbox jean

beamalsky:
## Overview

Add Jean to blackbox and drop beamalsky@gmail.com (expired; GPG Keychain shows I've renewed it successfully but I couldn't get blackbox to recognize that) in favor of beamalsky@datamade.us


jeancochrane:
@beamalsky I'm going to go ahead and merge this so I can keep moving, hope that's OK!

Issue #23, opened 2019-02-18
Survey tool: List view of collected data


reginafcompton:
The user should be able to view a list of all run surveys. 

This issue refers to simply setting up the view/route for collected data.

beamalsky:
@jeancochrane, what do you think we should call this route? https://justspaces.datamade.us/surveys/ is surveys that can be run, and I think we need something to distinguish that this view is for the data, not the forms. 

Some options: 

- https://justspaces.datamade.us/submitted <-- this is what fobi does
- https://justspaces.datamade.us/collecteddata
- https://justspaces.datamade.us/surveydata

beamalsky:
Another pattern that Fobi uses, which I don't think is very RESTful:

**https://justspaces.datamade.us/surveys/[Survey ID]** <-- this is where you run the survey
**https://justspaces.datamade.us/surveys/[Survey ID]/submitted** <-- this is where you look at the data you collected with the survey

beamalsky:
I'm imagining the collected data list view has a table of this form:

| Survey Title | Last submitted | 
| -----|------ |

When you click on a survey title, it brings you to a detail view for that survey, which will include ways to interact with individual submissions of that survey.

jeancochrane:
Hmm, this is a tricky question from a design perspective. Since "surveys" and "collected data" basically refer to the same entity in the context of a list view -- "collected data" is really just referring to "surveys that have been run," as I understand it -- part of me feels like the right idiom would be to include a boolean query parameter `?submitted`. This would imply that the table structure would be the same for the "survey" view as it would be for the "collected data" view, in that the parameter would only adjust the query that we run to retrieve the list of surveys.

That being said, it seems like maybe we don't actually want the "survey" table and the "collected data" table to have the same structure, in which case "surveys" and "collected data" are sort of referring to different things. In that case my preference would probably be for `/surveys/submitted/` for the list view and `/surveys/submitted/<survey_id>` for the detail view, which feels like it preserves the fact that "collected data" are essentially a special case of "surveys" while also insisting on the major distinction between them by using separate routes.

What do you think about that @beamalsky ?

beamalsky:
It is tricky, thanks for this! My sense is that we want editing/running a survey and interacting with that survey's submitted data to feel like pretty distinct actions and allow the likely possibility that those tables will have different structures. It makes more sense to me to follow your `/surveys/submitted/` and `/surveys/submitted/<survey_id>` suggestion.

[...]

Issue #61, opened 2019-03-18
Detail view of collected data: Initiate and implement a basic table


reginafcompton:
This issue should address:

* setting up view/route for collected data detail view
* presenting survey data in a table

reginafcompton:
@beamalsky - we cannot display data in a table, until we implement a strategy for data handling, right? If so, let's limit the scope of this issue to initiating view/route for the detail view.

beamalsky:
That's right @reginafcompton !

beamalsky:
Closed by #76

PR #71, opened 2019-04-08
Collected data list route and view

beamalsky:
## Overview

This PR creates a route and view for a simple list view of collected data, as described in #23 

### Notes

To do:

- [x] Figure out consistent naming for collected data views. @jeancochrane, I'd love your input on this. See: https://github.com/datamade/just-spaces/issues/23#issuecomment-480892552
- [x] Group by survey title. See https://github.com/datamade/just-spaces/issues/23#issuecomment-480905922

## Testing Instructions

 * Start with the db in just-spaces.dump
* Run the site locally and submit a survey! 
* Navigate to http://127.0.0.1:8000/collected-data/

Connects #23


beamalsky:
Hm, I didn't have any issues applying that new PLDP migration. I've added a new db dump with it, could you try this one out? 

And ee I deleted some data handler urls I didn't think were being used anywhere, but I was wrong! Adding them back should resolve your last comment.

Writing a test here is a great idea—I'm not in that habit and would like to be! Let me give it a try now.

jeancochrane:
Great, new DB dump worked for me! And the missing route is back now. Let me know when you're ready for me to look at the test.

tests/conftest.py
```diff
@@ -102,3 +105,43 @@ def form_element(db, form_entry):
...
+        total=5,
+        survey=survey,
+    )
+
+    return survey_row
+
+
+@pytest.fixture
+@pytest.mark.django_db
+def survey_component(db, row):
```
jeancochrane: Are `survey_row` and `survey_component` being used in the test at all, or are you setting them up for the future?

beamalsky: Just setting them up for the future! I'll need them to test the next issue (#61)

Issue #73, opened 2019-04-08
Data handler name error


beamalsky:
When I submit a Clark Park Movie Night survey without all of the elements filled out, it shows a successful submission page but this error in the terminal:

```
[08/Apr/2019 20:41:51] "GET /surveys/view/movies-in-clark-park/ HTTP/1.1" 200 47811
Error in class CollectDataPlugin. Details: 'name'. Full trace: Traceback (most recent call last):
  File "/Users/beamalsky/.virtualenvs/just-spaces/src/django-fobi/src/fobi/base.py", line 1835, in _run
    form_element_entries)
  File "/Users/beamalsky/Desktop/just-spaces/fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py", line 57, in run
    name = json_plugin_data['name']
KeyError: 'name'
```

Something's not quite right with the data handler. Investigate this!

PR #74, opened 2019-04-09
Bug fix for data handler name error (issue 73)

beamalsky:
## Overview

A fix for #73. This was simple! I hadn't realized that Fobi treats help text (`FormElementEntry with plugin_uid='context_text'`) as elements with no names, which needed to be handled in the data handler. I did that and added a test to make this easier to identify if it happens again. 

I also broke my tests into two separate files, which seems useful as they grow. The data handler test is giant but I'm not sure how to break it into smaller pieces without rerunning a bunch of code—@jeancochrane, I'd love to hear how you'd approach this. Are smaller tests worth the redundancy?

## Testing Instructions

 * Run pytest!

Connects #73


fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
```diff
@@ -54,21 +54,27 @@ def run(self, form_entry, request, form, form_element_entries=None):
...
+                    name=name,
+                    label=label,
+                    type=type,
+                    position=position,
+                    saved_data=saved_data
+                )
+
+            # Help text elements have no name and should not be saved
+            except KeyError:
+                pass
```
jeancochrane: At what point do we expect this `KeyError` will get raised? On the assignment `name = json_plugin_data['name']`? It's best practice to keep `try/except` blocks scoped as tightly as possible to the blocks that you expect to raise errors -- in this case:

```python
try:
   name = json_plugin_data['name']
except KeyError:
   continue  # Skip to the next object in the enclosing loop
```

I'm also wondering if there's any more specific way that we can check that the field is for help text other than to see whether it raises a `KeyError` here. Seems like we're assuming that form fields will be missing the `name` attribute if and only if they're help text components, and I don't know how safe of an assumption that is. Does Fobi expose any identifying information about help text components that we could use to perform a more specific check? E.g. is the `plugin_uid` attribute passed into the form at all?

tests/test_surveys_submitted.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+from django.urls import reverse
+
+
+@pytest.mark.django_db
+def test_survey_submitted_list(user, survey, client, form_entry):
```
jeancochrane: What do you think about making this file `tests/test_views.py`? `tests/test_surveys_submitted.py` makes sense if we anticipate we'll have a lot of tests for submitted surveys in particular, otherwise this test might fit better in a category that encompasses "all tests checking basic attributes of views".

tests/test_data_handling.py
```diff
@@ -10,7 +9,8 @@
 
 
 @pytest.mark.django_db
-def test_data_handler(form_entry, location, study, form_element, mocker):
+def test_data_handler(form_entry, location, study, form_element,
+                      form_element_help_text, mocker):
```
jeancochrane: I see what you're saying about this test being long. In this case I think it's fine, since what you're trying to do is save one type of object and make sure that a bunch of different attributes got saved correctly. If the test were saving different types of objects and testing their attributes, it would make sense to me to split it out into multiple tests, but in this case I think it's better to sacrifice brevity of the test in order to minimize database access and scope the test to one specific unit of functionality.

In general I tend to be less strict about code quality concerns (like DRY and keeping functions short) when writing tests. Tests are often verbose, which I think is OK since they're not really "methods" in a traditional sense -- they're usually more declarative than imperative, and I typically try to write them keeping in mind that the person reading them will most likely be skimming to try to debug a failure, not reuse functionality.

beamalsky: Very helpful, thanks for explaining!

tests/test_surveys_submitted.py
beamalsky: love it

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
beamalsky: I see your reasoning! The only plugin data that `content_text` pass into the form is a `text` key, but that's the only element that has it. I've switched to checking for that!

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
```diff
@@ -54,7 +54,11 @@ def run(self, form_entry, request, form, form_element_entries=None):
             plugin_data = element.plugin_data
             json_plugin_data = json.loads(plugin_data)
 
-            try:
+            # Check if an element is help text. If it is, skip it
+            if 'text' in json_plugin_data.keys():
+                pass
```
jeancochrane: I think a `continue` would be more explicit here, to indicate that we want to skip this form element entirely. Any reason you prefer `pass`?

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
```diff
@@ -54,21 +54,29 @@ def run(self, form_entry, request, form, form_element_entries=None):
...
+            else:
+                name = json_plugin_data['name']
+
+                label = json_plugin_data['label']
+                type = element.plugin_uid
+                position = element.position
+
+                saved_data = form.cleaned_data[name]
+
+                print(saved_data)
```
jeancochrane: Stray `print` statement:

```suggestion
```

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
beamalsky: Nope continue makes more sense!

Issue #72, opened 2019-04-08
New element creation error


beamalsky:
When I create a new survey locally and try to add a new element to it, I get this error: 

```
IntegrityError at /elements/create/4/float/
duplicate key value violates unique constraint "fobi_formelemententry_pkey"
DETAIL:  Key (id)=(2) already exists.
```

That's not good, let's fix it! This is with a database restored from just-spaces.dump.

beamalsky:
Fixed by #75

PR #75, opened 2019-04-09
New dump with id sequences (issue 72)

beamalsky:
## Overview

Addresses #72. I followed the method recommended [here](https://hcmc.uvic.ca/blogs/index.php/how_to_fix_postgresql_error_duplicate_ke?blog=22) and ran 

`SELECT setval('the_primary_key_sequence', (SELECT MAX(the_primary_key) FROM the_table)+1);` 

on all tables with sequences. I then created a new dump with

`pg_dump -Fc -U postgres just-spaces > just-spaces.dump`

I tested it by going into the database and seeing if the id sequences were retained:

```bash
$ psql just-spaces
=# SELECT nextval('fobi_formelemententry_id_seq');
```

They were!

```
 nextval
---------
      60
(1 row)
```

## Testing Instructions

* Drop the database: `dropdb just-spaces`
* Recreate the database: `pg_restore -C -j4 --no-owner just-spaces.dump | psql`
* Run the server: `python manage.py runserver`
* Login is the same as the staging site as testuser
* Create a new survey
* Add an element to it. Does it let you save it or do you get an error?

Connects #72


beamalsky:
Huh, that Postgres incompatibility isn't ideal but is good to know. And good call! I'll do that and add the password to the DataMade Lastpass

PR #76, opened 2019-04-09
Surveys submitted detail

beamalsky:
## Overview

This PR implements a basic detail view with a side scrolling DataTable, following the pattern of the sample data in Gehl's [PLDP repo](https://github.com/gehl-institute/pldp/tree/master/Sample%20Data). It also includes merged bug fixes from #72 and #73.

## Testing Instructions

* If a `just-spaces` database already exists, drop it:  `dropdb just-spaces`
* Recreate it: `pg_restore -C -j4 --no-owner just-spaces.dump | psql`
* Run the server: python manage.py runserver
* Login is as testuser (pw in DataMade Lastpass)
* Click "Run survey" in the sidebar. Submit some data!
* Click "View collected data" in the sidebar and select the survey you just ran.
* Check if the data you just submitted shows up!

Connects #61


surveys/views.py
```diff
@@ -43,12 +44,31 @@ class SurveySubmittedList(TemplateView):
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        context['form_entry'] = FormEntry.objects.get(id=context['form_entry_id'])
+        context['surveys_submitted'] = Survey.objects.filter(form_id=context['form_entry_id'])
+
+        for survey in context['surveys_submitted']:
+            row = SurveyRow.objects.get(survey=survey)
+            survey.row_total = row.total
+            survey.components = SurveyComponent.objects.filter(row=row).order_by('position')
```
jeancochrane: One alternative pattern that could be useful here would be to move these queries to property methods on the `Survey` model. That way if you wind up wanting to get the row or components for a Survey often, you can treat them like properties on the model instead of having to remember the queries for them each time. Something like:

```python
class Survey(db.model):
...
    @property
    def row(self):
        ...

    @property
    def components(self):
    ...
```

I think this is an optional step at this point, but it's something to keep in mind as you continue to flesh out the relationship between `Survey`s, `SurveyRow`s, and `SurveyComponents`.

beamalsky: Ah nice I'd like to get more comfortable with this pattern! Here's a PR in the `django-pldp` repo to get it working: https://github.com/datamade/django-pldp/pull/10

beamalsky:
@jeancochrane this passed checks with the new PLDP properties! still ok to merge by you?

jeancochrane:
@beamalsky Merge away!

Issue #54, opened 2019-03-13
Published vs. unpublished survey states


beamalsky:
In order to maintain continuity between surveys, once a survey form has been run once it should no longer be editable. We plan on restricting this by having distinct published (runnable) and unpublished (editable) states. 

It will be possible to published an unpublished survey, but not to go back to unpublished once a survey has been published. This was run by UCD in our wireframes!

Issue #46, opened 2019-03-11
Implement "two-track" survey creation


reginafcompton:
Intercept and observational surveys have different requirements. These differences emerge, particularly, in how the surveys handle "counted" data, e.g., "how many people are riding bikes in the park?" vs. "did you ride a bike in the park?" 

Currently, the survey builder offers both types of questions (single vs. multiple count). 

<img width="290" alt="Screen Shot 2019-03-11 at 2 29 32 PM" src="https://user-images.githubusercontent.com/13078679/54148233-40db8200-440a-11e9-8019-6216e81592bd.png">

This can be confusing or "clunky." Instead, the form elements should depend on the type of survey. 

**On the `create` view, let's give users the option to chose one of two "tracks": observational or intercept.**

****

**Important note!** The PLDP handles observational data, but not intercept. For now, users who follow the observational track will also adhere to a strict implementation of PLDP. Alternatively, the intercept surveys will have greater flexible, since intercept surveys aim to capture fine-grained detail, gather opinions, and record non-observable facts.

reginafcompton:
We want to clearly communicate to the user the benefits and disadvantages of departing from the Protocol (e.g., departures from PLDP compromise the share-ability and comparability of survey data). 

For users who select the "intercept" track, we might want some clarifying text, like, “intercept surveys do not follow PLDP....which means...."?

beamalsky:
First steps that @jeancochrane and I determined for this:

Subclass the `django-fobi` FormEntry model in this repo to include new metadata like a `published` boolean (related to #63) and then into two subsubclasses for Intercept and Observational studies. Determine how much this breaks Fobi. Would it be possible to write our own views for the survey creation form, and then keep the Fobi views for the survey element editing functionality? If the answer is no and we're making major changes to Fobi, consider bringing it into the Just Spaces repo instead of maintaining it as a separate fork at https://github.com/datamade/django-fobi

PR #83, opened 2019-04-10
Start of a subclassed SurveyFormEntry

beamalsky:
## Overview

This PR creates a custom SurveyFormEntry subclass to use on top of Fobi's FormEntry class. It includes published, location, and study fields, and plays nicely with the edit survey views we're currently using in Fobi. I've also updated tests to reflect this model change. This is a first step in implementing two-track survey creation (#46) and in handling published and unpublished survey states (#54).

## Testing Instructions

* Create a new survey
* Run it
* Review submitted data. Is it all there?


surveys/templates/survey_submitted_detail.html
```diff
@@ -38,7 +38,7 @@ <h2 class="card-title">{{ form_entry.name }}</h2>
 <script>
 $(document).ready(function () {
   $('#survey-submitted-detail').DataTable({
-    "scrollX": true
+    // "scrollX": true,
```
jeancochrane: What was going on here?

surveys/views.py
```diff
@@ -48,7 +69,11 @@ def get_context_data(self, **kwargs):
         context['surveys_submitted'] = surveys.order_by('form_id', '-time_stop').distinct('form_id')
 
         for survey in context['surveys_submitted']:
-            survey.form_title = FormEntry.objects.get(id=survey.form_id)
+            try:
+                survey_form_entry = SurveyFormEntry.objects.get(formentry_ptr=survey.form_id)
+                survey.form_title = survey_form_entry.name
+            except SurveyFormEntry.DoesNotExist:
+                survey.form_title = "[Deleted Survey]"
```
jeancochrane: It seems like this logic represents defensive programming in the case that the user has deleted the referenced `SurveyFormEntry`, is that right? I wonder if there's a safer way that we can handle this case. My instinct would be to enforce `ON DELETE CASCADE` so that when a `SurveyFormEntry` gets deleted all of the `Survey`s that reference it will also get deleted, but I understand that we can't enforce that at the ORM level in `django-pldp` because the `Survey.form_id` attribute shouldn't be a foreign key to a separate app.

I wonder if we could [add a custom constraint](https://www.fusionbox.com/blog/detail/custom-database-constraints-in-django/594/) to the Survey table with a migration in the `surveys` app? The constraint I'm imagining would be something like:

```sql
ALTER TABLE pdlp_survey
ADD CONSTRAINT form FOREIGN KEY (form_id)
REFERENCES surveys_surveyform (id) ON DELETE CASCADE
```

I'm not really sure if this would work since I don't know if we can reliably say that the `pldp_survey` table will exist before the `surveys_surveyform` table in the app (I would guess that the order of the apps in `INSTALLED_APPS` would determine this). Still, I think it's worth a shot.

surveys/templates/survey_submitted_detail.html
beamalsky: DataTables' ScrollX styling was wonky certain kinds of submitted data. Got it working with this fix! https://stackoverflow.com/questions/45136218/datatables-scrollx-causing-squashed-header?rq=1

surveys/views.py
beamalsky: I opened issue #84 to figure this out. @jeancochrane, do you want to add your approach there?

jeancochrane: Done!

surveys/templates/survey_submitted_detail.html
jeancochrane: Nice! Do you mind adding a quick comment to this code describing the problem and linking to this SO answer? (You can use the "share" URL for brevity: https://stackoverflow.com/a/47800221) I know in 6 months I'm not going to remember why we added these options.

Issue #32, opened 2019-03-04
RESTFUL routes with query params


reginafcompton:
Related to #26.

We will update the routes for the for the survey list views, when we have a clear understanding of what makes a survey "editable" and what makes a survey "run-able." 

### Surveys – ListView
RESTful urls should identify a resource. The url should fetch a particular kind of resource (e.g., published vs. unpublished) with query params. [(See point 8.)](https://medium.com/hashmapinc/rest-good-practices-for-api-design-881439796dc9)

List view of surveys that can be edited (replaces “edit-survey”)
`/surveys?state=unpublished`

List view of surveys that can be run (replaces “run-survey”)
`/surveys?state=published`

With this, views.py should have a SurveyListView, rather than EditSurvey and RunSurvey. The view should check for query params, and return a filtered Survey queryset.

PR #85, opened 2019-04-15
Add publish view

beamalsky:
## Overview

This PR adds a view, template, and url for publishing surveys. Once this is approved I can move forward with #32 and improving the `survey-list` template to better reflect the difference between published and unpublished surveys.

It'll be nice to have tests on this, but I think that makes more sense to implement once I'm also working on #32 

## Testing Instructions

* Create a new survey
* Go to the survey list view
* Do you see a button under the "Published" column?
* Click it. Does it prompt you on whether or not you want to publish the survey?
* Submit that. Does the survey no longer have a "Publish" button?

Connects #54

### DIFF
--- a/surveys/templates/base.html
+++ b/surveys/templates/base.html
@@ -1,4 +1,4 @@
-{% load staticfiles %}
+{% load static %}
 
 <!doctype html>
 <html lang="en">
--- a/surveys/templates/survey_publish.html
+++ b/surveys/templates/survey_publish.html
@@ -1,0 +1,12 @@
+{% extends 'base.html' %}
+{% load static %}
+
+{% block content %}
+  <h2>Publish survey</h2>
+  <form method="POST" class="post-form">{% csrf_token %}
+     <div>Are you sure you want to publish <b>{{survey.name}}</b>?</div>
+     <div>Surveys must be published before they can be run, but cannot be edited
+       once they've been published.</div>
+     <button type="submit" class="save btn btn-default">Publish</button>
+  </form>
+{% endblock %}
--- a/surveys/templates/survey_publish.html
+++ b/surveys/templates/survey_publish.html
@@ -4,9 +4,9 @@
 {% block content %}
   <h2>Publish survey</h2>
   <form method="POST" class="post-form">{% csrf_token %}
-     <div>Are you sure you want to publish <b>{{survey.name}}</b>?</div>
-     <div>Surveys must be published before they can be run, but cannot be edited
-       once they've been published.</div>
+     <p>Are you sure you want to publish <b>{{survey.name}}</b>?</p>
+     <p>Surveys must be published before they can be run, but cannot be edited
+       once they've been published.</p>
      <button type="submit" class="save btn btn-default">Publish</button>
   </form>
 {% endblock %}
--- a/surveys/views.py
+++ b/surveys/views.py
@@ -1,4 +1,4 @@
-from django.views.generic import TemplateView, ListView
+from django.views.generic import TemplateView, ListView, View
 from django.views.generic.edit import CreateView, FormView
 from django.shortcuts import render, redirect
 from django.http import HttpResponseRedirect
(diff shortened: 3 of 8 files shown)


beamalsky:
Hm, I thought Travis wasn't going for some reason related to Civic House's network problems, but that doesn't seem to be right. @jeancochrane do you know if there's some Travis setting that says to only run on branches based on master? I'm not sure what's up here.

surveys/views.py
```diff
@@ -53,6 +53,23 @@ def form_valid(self, form):
...
+        context['survey'] = SurveyFormEntry.objects.get(id=context['form_entry_id'])
+
+        return context
+
+    def post(self, request, **kwargs):
+        context = self.get_context_data(**kwargs)
+        context['survey'].published = True
+        context['survey'].save()
+
+        return redirect('survey-list')
```
jeancochrane: When I try to publish a survey I get this error:

```
Internal Server Error: /surveys/publish/4/
Traceback (most recent call last):
  [...]
  File "/home/datamade/.virtualenvs/just-spaces/lib/python3.6/site-packages/django/urls/resolvers.py", line 622, in _reverse_with_prefix
    raise NoReverseMatch(msg)
django.urls.exceptions.NoReverseMatch: Reverse for 'survey-list' not found. 'survey-list' is not a valid view function or pattern name.
```

I think the right URL name might be `surveys-list`? In any case, this seems like a good opportunity for a test that would `POST` to publish a survey and confirm that the response returns 200.

surveys/templates/survey_publish.html
```diff
@@ -0,0 +1,12 @@
+{% extends 'base.html' %}
+{% load static %}
+
+{% block content %}
+  <h2>Publish survey</h2>
+  <form method="POST" class="post-form">{% csrf_token %}
+     <div>Are you sure you want to publish <b>{{survey.name}}</b>?</div>
```
jeancochrane: Any reason for these to be `<div>`s instead of `<p>`s? (And for them to be nested inside the form?)

beamalsky: None of the forms have final styles, I figured that should be a separate issue!

beamalsky: I can switch these to `<p>`s though

surveys/views.py
beamalsky: Oops sorry about that, you're right about `surveys-list`! I've added a test in the last commit to test survey publishing, but I can't figure out why the last assertion isn't passing. 

Could we pair on this tomorrow?

jeancochrane: Sure thing!

jeancochrane: Took a quick look at this because I had an idea of what might be happening. I think you need to refresh the `survey_form_entry` object from the DB in order to see the updated attribute. Docs: https://docs.djangoproject.com/en/dev/ref/models/instances/#django.db.models.Model.refresh_from_db

I was able to confirm that `survey_form_entry.published` was `True` after calling `survey_form_entry.refresh_from_db()`.

beamalsky: Wow interesting!! Thanks, this worked.

surveys/templates/base.html
```diff
@@ -1,4 +1,4 @@
-{% load staticfiles %}
+{% load static %}
```
jeancochrane: Good catch! I noticed this but forgot to mention it. Where possible, fixing deprecation warnings is a good move.

tests/conftest.py
```diff
@@ -135,7 +136,7 @@ def form_element_help_text(db, survey_form_entry):
 @pytest.mark.django_db
 def survey(db, location, study):
     survey = Survey.objects.create(
-        time_stop=datetime.now(tz=None),
+        time_stop=datetime.now(tz=get_current_timezone()),
```
jeancochrane: Another good call.

beamalsky: Pytest pointed it out as a warning! Thank you pytest. 🙏🏻

PR #86, opened 2019-04-16
Published state

### LINKED ISSUE #32: RESTFUL routes with query params

beamalsky:
## Overview

This PR closes both #32 and #54 by checking for a `published` value in the SurveyList view. I feel like there might be a more elegant way to modify `urls.py` to handle this!

## Testing Instructions

* Create a couple new surveys. Publish at least one. Is the interface reasonably intuitive?

Connects #32 and #54

### DIFF
--- a/surveys/templates/partials/sidebar.html
+++ b/surveys/templates/partials/sidebar.html
@@ -26,13 +26,13 @@
                 </ul>
             </li>
             <li>
-                <a href="{% url 'surveys-list' %}">
+                <a href="{% url 'surveys-list' %}?published=f">
                     <i class="fas fa-pencil-alt icon edit"></i>
                     Edit survey
                 </a>
             </li>
             <li>
-                <a href="{% url 'surveys-list' %}">
+                <a href="{% url 'surveys-list' %}?published=t">
                     <i class="fas fa-clipboard-list icon run"></i>
                     Run survey
                 </a>
--- a/surveys/templates/partials/sidebar.html
+++ b/surveys/templates/partials/sidebar.html
@@ -26,13 +26,13 @@
                 </ul>
             </li>
             <li>
-                <a href="{% url 'surveys-list' %}?published=f">
+                <a href="{% url 'surveys-list-edit' %}">
                     <i class="fas fa-pencil-alt icon edit"></i>
                     Edit survey
                 </a>
             </li>
             <li>
-                <a href="{% url 'surveys-list' %}?published=t">
+                <a href="{% url 'surveys-list-run' %}">
                     <i class="fas fa-clipboard-list icon run"></i>
                     Run survey
                 </a>
--- a/surveys/views.py
+++ b/surveys/views.py
@@ -1,4 +1,4 @@
-from django.views.generic import TemplateView, ListView, View
+from django.views.generic import TemplateView, ListView
 from django.views.generic.edit import CreateView, FormView
 from django.shortcuts import render, redirect
 from django.http import HttpResponseRedirect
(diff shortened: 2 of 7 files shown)


surveys/templates/partials/sidebar.html
```diff
@@ -26,13 +26,13 @@ <h3><b>Just Spaces</b></h3>
                 </ul>
             </li>
             <li>
-                <a href="{% url 'surveys-list' %}">
+                <a href="{% url 'surveys-list' %}?published=f">
```
jeancochrane: I typically prefer to use `true` and `false` for boolean values instead of `t` and `f` (or `0` and `1`). There's no functional difference, it just makes the URL easier for me to read. Would be curious to hear your thoughts if you feel differently. Some discussion on SO: https://stackoverflow.com/q/21254159

surveys/templates/partials/sidebar.html
```diff
@@ -26,13 +26,13 @@ <h3><b>Just Spaces</b></h3>
...
                     <i class="fas fa-pencil-alt icon edit"></i>
                     Edit survey
                 </a>
             </li>
             <li>
-                <a href="{% url 'surveys-list' %}">
+                <a href="{% url 'surveys-list' %}?published=t">
```
jeancochrane: What should get returned if the user visits `/surveys/` with no `published` param? I typically think of query parameters as offering optional filtering on top of a base view, but in this case it seems like the parameter is required in order to determine whether the view is for running or editing surveys.

If it makes sense to have a `/surveys/` endpoint that shows both published and unpublished surveys (something like `All Surveys`), then I think we should implement that endpoint as part of this PR. If that doesn't make sense, then I think we should reconsider the `published` param and instead refactor the URL hierarchy, e.g.:

- `/surveys/?published=t` -> `/surveys/run/`
- `/surveys/?published=f` -> `/surveys/edit/`

What do you think?

surveys/views.py
```diff
@@ -67,18 +67,31 @@ def post(self, request, **kwargs):
...
     model = SurveyFormEntry
     template_name = "survey_list.html"
     context_object_name = 'surveys'
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context['published'] = self.request.GET.get('published', 'f')
-        context['surveys'] = context['surveys'].filter(published=context['published'])
+        context['surveys'] = context['surveys'].filter(published=False)
+        context['published'] = False
```
jeancochrane: Note that we could also accomplish this filtering in the [`ListView.get_queryset()`](https://docs.djangoproject.com/en/dev/ref/class-based-views/mixins-multiple-object/#django.views.generic.list.MultipleObjectMixin.get_queryset) method. Both approaches work fine, but if at some point this method starts to get unweildy, you could factor out the portions that alter the `surveys` object into `get_queryset()`.

PR #87, opened 2019-04-16
Survey metadata

beamalsky:
## Overview

Connects #63. Creates elements for survey metadata. Related issues that came out of this but should be addressed separately: #88; #91; #92; #93 

### To Do 

Add fields on survey model:

- [x] location
- [x] study

Add form elements:
- [x] total
- [x] time_start
- [x] time_stop
- [x] time_character
- [x] representation
- [x] microclimate
- [x] temperature_c
- [x] method

Other things:

- [x] We'll need observational surveys to require a `Total` form element. And maybe `time_start`, `time_stop`, `method`, and `representation`. Open a new issue to handle these. (update: issue #88)
- [x] Get data handler working with all of this
- [x] Figure out relationship between time_stop and submission time (update: issue #91)

## Testing Instructions

* Create a new survey
* Add some of the metadata fields
* Publish your survey
* Submit it. Do the fields save as you'd expect on the survey model?

Connects #63


fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
```diff
@@ -85,4 +78,44 @@ def plugin_data_repr(self):
     return self.data.__dict__
 
 
+def get_element_info(self, form_element):
```
jeancochrane: Any reason these functions can't be methods on the `CollectDataPlugin` class?

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
```diff
@@ -85,4 +78,44 @@ def plugin_data_repr(self):
...
+
+
+def get_saved_data(self, plugin_uid, default_value=''):
+    try:
+        form_element = FormElementEntry.objects.get(form_entry_id=self.form_id, plugin_uid=plugin_uid)
+        saved_data = get_element_data(self, form_element)
+        if (plugin_uid in ['time_start', 'time_stop']) and saved_data:
+            saved_data = datetime.combine(self.today, saved_data).replace(tzinfo=self.timezone)
+    except FormElementEntry.DoesNotExist:
+        saved_data = default_value
```
jeancochrane: You could scope this exception handling to the specific call that might raise an exception by using `try`/`except`/`else`:

```python
    try:
        form_element = FormElementEntry.objects.get(form_entry_id=self.form_id, plugin_uid=plugin_uid)
    except FormElementEntry.DoesNotExist:
        saved_data = default_value
    else:
        saved_data = get_element_data(self, form_element)
        if (plugin_uid in ['time_start', 'time_stop']) and saved_data:
            saved_data = datetime.combine(self.today, saved_data).replace(tzinfo=self.timezone)
```

justspaces/settings.py
```diff
@@ -130,7 +133,7 @@
 
 LANGUAGE_CODE = 'en-us'
 
-TIME_ZONE = 'UTC'
+TIME_ZONE = 'America/Chicago'
```
jeancochrane: Do we want `America/New_York` here?

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [tool:pytest]
-DJANGO_SETTINGS_MODULE = tests.test_config
+DJANGO_SETTINGS_MODULE = justspaces.settings
```
jeancochrane: Is this change just because the test settings weren't changing anything? Do the tests not use a separate DB?

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
beamalsky: They can!

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
beamalsky: Nice 😎

setup.cfg
beamalsky: Yep, they were the same and it seemed like an unnecessary hassle to have to change settings in two places.

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
```diff
@@ -69,53 +69,51 @@ def run(self, form_entry, request, form, form_element_entries=None):
...
+                saved_data = datetime.combine(self.today, saved_data).replace(tzinfo=self.timezone)
 
-def get_element_data(self, form_element):
-    element_info = get_element_info(self, form_element)
-    saved_data = element_info['saved_data']
+        return saved_data
 
-    return saved_data
 
+def plugin_data_repr(self):
```
jeancochrane: I would expect this to be a method on the class too, but it looks like it's dedented. Is there a reason for that?

beamalsky: This is a holdover from Fobi! I don't see a need to mess with it, since it works and I'm not sure where else it's used in the package.

jeancochrane: Sounds great, thanks for clarifying.

Issue #48, opened 2019-03-11
Make a plan for handling user permissions


reginafcompton:
Different users will have different levels of access to the site. We need to concretize this hierarchy before implementation. 

- [ ] Email UCD, and remind them to think about this (as a follow up to our March 11 meeting)
- [ ] Perhaps, have another phone call about this.
- [ ] **Put into writing** the exact specifications of user permissions – both UCD and DataMade need to agree on these specifications.

beamalsky:
For now, everyone is a superuser. This will need to change! I've turned off Fobi's permissioning system because I think we're better served building our own.

beamalsky:
I propose we set up either 2 or 3 types of users:

1. **Users:** The lowest level of permissions. Users can run surveys to which they've been granted access.
2. **Admins:** The middle level of permissions. Admins can run surveys as well as create, edit, and delete Studies, Study Areas, Surveys, and Locations. They can grant Users access to surveys. Admins can also view collected data and design data visualizations on the collected data pages.
3. **Superusers:** The highest level of permissions. In addition to the permissions of Users and Admins, Superusers can create, edit, and delete Agencies and other accounts.

New accounts are created as users, and can be promoted to Admins or Superusers by Superusers. I'm still thinking on this, and it might make sense to squash 2 and 3 and pitch UCD with just two levels—Users and Admins—and figure out if there's a need for more differentiated roles from there.

beamalsky:
Plan approved by UCD! See #173

PR #94, opened 2019-04-17
Two track surveys

### LINKED ISSUE #46: Implement "two-track" survey creation

beamalsky:
## Overview

This is the last bit of work needed to close #46! https://github.com/datamade/django-fobi/pull/2 will need to be merged before it passes checks. 

### To do 

- [x] Write a test or two!

## Testing Instructions

* Create a new survey with type Observational
* When you go to add a new question, do you see the "Observational" category?
* Create a new survey with type Intercept
* When you go to add a new question, do you NOT see the "Observational" category?

Connects #46


justspaces/settings.py
```diff
@@ -151,4 +151,4 @@
 
 FOBI_DEFAULT_THEME = 'foundation5'
 FOBI_THEME_FOOTER_TEXT = gettext.gettext('')
-FOBI_RESTRICT_PLUGIN_ACCESS = False
+FOBI_RESTRICT_PLUGIN_ACCESS = True
```
jeancochrane: I'm not totally confident I get what this is doing -- from the docs:

> FOBI_RESTRICT_PLUGIN_ACCESS (bool): If set to True, (Django) permission system for dash plugins is enabled. Defaults to True. Setting this to False makes all plugins available for all users.

Does this mean that some users won't have access to some plugins, or that plugins can only be used on the group of which they're a part?

beamalsky: I changed this when I was trying to figure out how to approach this issue and forgot to change it back—my bad! I think we shouldn't use Fobi's permissions system but should instead build our own once we get there.

beamalsky:
@jeancochrane I've added a column in the list view to indicate if a survey is observational or intercept! Now that I've merged these [changes](https://github.com/datamade/django-fobi/pull/2) in `django-fobi`, you should see the right filtered categories in the form element dropdown. Could you give it another shot?

beamalsky:
For these new tests to work, I made the user fixture a superuser. We'll probably want to change that once we address #48

tests/test_views.py
```diff
@@ -26,6 +26,33 @@ def test_survey_list_run(client, user, survey_form_entry):
...
+@pytest.mark.django_db
+def test_survey_edit_intercept(client, user, survey_form_entry):
+    client.force_login(user)
+    url = reverse('fobi.edit_form_entry', kwargs={'form_entry_id': survey_form_entry.id})
+    response = client.get(url)
+
+    plugins = response.context['user_form_element_plugins']
+
+    print(plugins)
+    print(survey_form_entry.type)
```
jeancochrane: Reminder to delete these print statements before bringing this in.

tests/test_views.py
```diff
@@ -26,6 +26,33 @@ def test_survey_list_run(client, user, survey_form_entry):
...
+@pytest.mark.django_db
+def test_survey_edit_observational(client, user, survey_form_entry_observational):
+    client.force_login(user)
+    url = reverse('fobi.edit_form_entry', kwargs={'form_entry_id': survey_form_entry_observational.id})
+    response = client.get(url)
+
+    plugins = response.context['user_form_element_plugins']
+
+    assert response.status_code == 200
+    assert len(plugins) == 4
```
jeancochrane: Nice tests!

Issue #89, opened 2019-04-17
Previewing unpublished surveys


beamalsky:
Currently, the only way to see what a survey looks like to the survey-taker is to publish it. That can't be undone, and I think a preview function for unpublished surveys would be pretty useful. It should let you see what a survey looks like but not submit data.

PR #107, opened 2019-04-22
Preview functionality

beamalsky:
## Overview

Adds ability to preview unpublished surveys. https://github.com/datamade/django-fobi/pull/3 should be reviewed and merged before this PR.

Success messages will need some work in order to be customizable and indicate when forms are in preview mode. I've opened an issue for that: #108 

## To do

- [x] Once the `django-fobi` PR linked above is merged: write a test!

## Testing Instructions

* Create a new survey
* Add at least one form element to it
* Go back to the list view and hit the "Preview button" 
* Does it say "Preview [Survey Name] at the top of that detail page?
* Submit your survey
* Check the collected data list view for your survey and make sure submitted data _isn't_ there

Connects #89


tests/test_views.py
```diff
@@ -51,20 +51,30 @@ def test_survey_edit_observational(client, user, survey_form_entry_observational
...
 @pytest.mark.django_db
-def test_survey_publish(client, survey_form_entry):
-    assert not survey_form_entry.published
+def test_survey_preview(client, user, survey_form_entry_observational):
+    client.force_login(user)
+    url = reverse('fobi.view_form_entry', kwargs={'form_entry_slug': survey_form_entry_observational.slug})
+    response = client.get(url)
+
+    assert response.status_code == 200
+    assert not response.context['form_entry'].surveyformentry.published
```
jeancochrane: Perhaps we could add a check here to make sure that the title of the page is correct? Something like:

```python
assert ('Preview ' + survey_form_entry_observational.name) in response.content.decode('utf-8')
```

beamalsky: Yes! I wanted to do this but wasn't sure how to access the page title. Thanks for this tip.

Issue #105, opened 2019-04-19
Create common groupings of PLDP form questions


jeancochrane:
Related to #104, create common groupings of observational PLDP questions. For example, right now we recommend that the user create form elements for "male," "female," and "unknown" when observing the gender of participants, but we don't have a way to quickly allow the user to implement that pattern, and leaving it up to the user means that our ACS comparisons in #100 may not be complete if the survey designer doesn't design the question properly.

This might be a tricky code design task, since I doubt Fobi supports this feature out of the box.

beamalsky:
Closed by #117 !

PR #111, opened 2019-04-24
Rearrange custom form elements

beamalsky:
As a first step toward #105, I've rearranged the custom plugins into three folders: metadata, intercept, and observational. These correspond with the groups that'll show up on the frontend form element selection menu. Since this involves touching a lot of files, what do you think about merging this before I continue with the issue @jeancochrane?


beamalsky:
Nice!

PR #113, opened 2019-04-24
Grouped Observational Age plugin

beamalsky:
## Overview

This PR implements a custom PLDP age widget for observational studies as described in #105. It uses Django's [MultiValueField](https://docs.djangoproject.com/en/2.2/ref/forms/fields/#multivaluefield) and [MultiWidget](https://docs.djangoproject.com/en/2.2/ref/forms/widgets/#multiwidget), which work pretty smoothly EXCEPT I spent a while spinning my wheels on how to add labels to each subwidget—ie, the age ranges as seen here:

Screen Shot 2019-04-29 at 2 26 41 PM

I think these sub-labels are necessary, but I'm not pleased with my method right now as it involves editing rendered HTML. (See [here](https://github.com/datamade/just-spaces/blob/issue-105/fobi_custom/plugins/form_elements/fields/observational/widgets.py#L16-L25)). It works, but I wonder if it could be better. This is something that [`format_output`](https://cdf.9vo.lt/1.10/django.forms.widgets/MultiWidget.html#format_output) could have handled before it was deprecated.

This widget currently saves data as a list of tuples. For example, the image below would turn into the following list.

Screen Shot 2019-04-29 at 2 38 06 PM

`[('0-14', 4), ('15-24', 10), ('25-64', 100), ('65+', 17)]`

This is a starting point that will be straightforward to change. @jeancochrane, let's chat about how this data should be formatted in order to connect smoothly with your visualizations.

Once we've merged this PR I'll still need to create the other observational elements described by the PLDP. However, I've made `ObservationalField` and `ObservationalWidget` reusable so this PR should include the heaviest technical lift for #105.

## Testing Instructions

* Add an Age element to it. Can you change the detail level, and does the resulting element look like you'd expect?

Connects #105


fobi_custom/plugins/form_elements/fields/observational/widgets.py
```diff
@@ -0,0 +1,36 @@
+import re
+from django import forms
+
+
+class ObservationalWidget(forms.MultiWidget):
+    # template_name = 'observational_widget.html'
```
jeancochrane: Reminder to scrap this comment before merging.

```suggestion
```

tests/test_models.py
```diff
@@ -13,8 +13,17 @@ def test_user(user):
...
     assert saved_survey_form_entry.type == 'intercept'
+
+
+@pytest.mark.django_db
+def test_survey_form_entry_observational(client, user, survey_form_entry_observational, form_element_observational):
+    saved_survey_form_entry = SurveyFormEntry.objects.get(id=2)
+
+    assert saved_survey_form_entry.name == 'Sample Form Entry Observational'
+    assert not saved_survey_form_entry.published
+    assert saved_survey_form_entry.type == 'observational'
```
jeancochrane: Could we test rendering this form to confirm that there are the correct number of inputs?

fobi_custom/plugins/form_elements/fields/observational/widgets.py
```diff
@@ -0,0 +1,36 @@
...
+            self.widgets = self.widgets + [forms.NumberInput()]
+
+        super(ObservationalWidget, self).__init__(self.widgets, *args, **kwargs)
+
+    def render(self, name, value, attrs=None, renderer=None):
+        context = self.get_context(name, value, attrs)
+        rendered_html = self._render(self.template_name, context, renderer)
+
+        split_html = re.findall('\<input.*?\>', rendered_html)
+        processed_html = ['<label>' + self.get_choice(self.choices, input, split_html) + '</label>' + input for input in split_html]
```
jeancochrane: I can see why this was making you uneasy. Let's have a quick call to talk about alternatives.

fobi_custom/plugins/form_elements/fields/observational/widgets.py
```diff
@@ -0,0 +1,36 @@
...
+
+        final_html = ''.join(processed_html)
+
+        return final_html
+
+    def get_choice(self, choices, input, string):
+        index = string.index(input)
+        _, choice = choices[index]
+        return choice
+
```
jeancochrane: Stray linebreak here.

```suggestion
```

fobi_custom/plugins/form_elements/fields/observational/age/forms.py
```diff
@@ -2,66 +2,58 @@
...
+                                     "5-9<br />"
+                                     "10-14<br />"
+                                     "15-17<br />"
+                                     "18-24<br />"
+                                     "25-34<br />"
+                                     "35-44<br />"
+                                     "45-54<br />"
+                                     "55-64<br />"
+                                     "65-74<br />"
+                                     "75+<br /><br />")
```
jeancochrane: Are these choices initialized with the default value `0`? It occurs to me that in the case of `detailed` and `complex` groupings, the survey runner is probably not going to want to fill out every field with `0` that doesn't contain a category they saw.

fobi_custom/plugins/form_elements/fields/observational/fields.py
```diff
@@ -0,0 +1,14 @@
+from django import forms
+
+
+class ObservationalField(forms.MultiValueField):
+    def __init__(self, choices=None, *args, **kwargs):
+        fields = [forms.IntegerField(label="ldkfjs")] * len(choices)
```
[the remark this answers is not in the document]
beamalsky: Nope, just an old placeholder.

fobi_custom/plugins/form_elements/fields/observational/fields.py
```diff
@@ -0,0 +1,14 @@
...
+        self.choices = choices
+        super(ObservationalField, self).__init__(fields, *args, **kwargs)
+
+    def compress(self, data_list):
+        choices_cleaned = [choice[0] for choice in self.choices]
+        saved_data = list(zip(choices_cleaned, data_list))
```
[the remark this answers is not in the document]
beamalsky: Sounds good to me!

fobi_custom/plugins/form_elements/fields/observational/age/forms.py
beamalsky: They're initialized with an empty box that currently saves as "None". I think there's value in keeping that input blank before the survey runner puts in any data, so they can tell the difference between a recorded 0 and a box they haven't looked at yet, but also that it might make sense to save those as 0 if there's no reason on the DAD end to distinguish those 2 cases.

tests/test_models.py
beamalsky: Yes!

fobi_custom/plugins/form_elements/fields/observational/widgets.py
beamalsky: The custom form renderers have me stalled, and I'm not sure this is worth digging into more: https://docs.djangoproject.com/en/2.2/ref/forms/renderers/

@jeancochrane could you take a look?

I also realized I was wrong about collected data containing `<input` causing a problem—we'd only run into issues if one of our subwidget labels included the text `<input`. Those aren't currently exposed to user input (they're the `choices` [described in the PLDP](https://github.com/datamade/django-pldp/blob/master/pldp/forms.py)), but the protocol does recommend allowing users to add their own options for some form elements. If we keep this method of producing labels for subwidgets, I think that should be a separate issue.

fobi_custom/plugins/form_elements/fields/observational/age/forms.py
jeancochrane: Good point. Let's punt on this and return to it once we have a better sense of data requirements for the DAD.

fobi_custom/plugins/form_elements/fields/observational/widgets.py
jeancochrane: Thanks for taking a closer look at this. I'd be interested to dig deeper into it but I don't have high confidence that I'll find a solution that you haven't, and I don't want to hold up this PR.

I say let's merge this and update the widget rendering if I can find a better path (or if we run itno errors down the line). Does that sound OK to you @beamalsky?

beamalsky: Sounds good!

beamalsky: Opened an issue for this: #116

Issue #116, opened 2019-04-30
Consider custom choices for observational questions


beamalsky:
Some PLDP elements—like age—have choices that are exhaustive. Some—like mode and posture—have detailed lists of choices, but suggest letting users create their own. Consider implementing this in `just-spaces`.

PR #117, opened 2019-04-30
Add remaining observational survey questions

beamalsky:
## Overview

Connects #105 

This PR implements the remaining PLDP survey questions: `mode_observational`, `posture_observational`, `activities_observational`, and `objects_observational`.

We may want to come up with a way for users to select a subset of these category options and maybe even mix detail levels. I've opened a separate  issue for that: #116 

## Testing Instructions

* Add each new element to it. Can you change the detail level, and does the resulting element look like you'd expect?

### DIFF
--- a/fobi_custom/plugins/form_elements/fields/utils.py
+++ b/fobi_custom/plugins/form_elements/fields/utils.py
@@ -1,0 +1,6 @@
+def choices_to_help_text(choices):
+    help_text = ''
+    for _, choice in choices:
+        help_text += '- ' + choice + '<br />'
+
+    return help_text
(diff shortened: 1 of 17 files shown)


fobi_custom/plugins/form_elements/fields/utils.py
```diff
@@ -0,0 +1,6 @@
+def choices_to_help_text(choices):
+    help_text = ''
+    for _, choice in choices:
+        help_text += '- ' + choice + '<br />'
+
+    return help_text
```
jeancochrane: Nice job factoring this out!

PR #120, opened 2019-04-30
Basic chart handling in the data analysis designer

### LINKED ISSUE #95: Initialize basic chart handling functionality for DAD
Initialize the Data Analysis Designer with some basic chart handling functionality:

- Add, edit, and delete "chart areas" (no need for actual charts yet)
- Add, edit, and delete a short description for each chart area
- Rearrange charts
- Save charts
- Clear all charts

Don't worry about the actual charts themselves yet.

jeancochrane:
## Overview

This PR implements basic chart handling functionality in the data analysis designer, including:

- Add, edit, and delete "chart areas" (no actual charts yet)
- Add, edit, and delete a short description for each chart area
- Rearrange charts
- Save charts
- Clear all charts

## Testing Instructions

* Run migrations: `python manage.py migrate`
* Start the dev server
* Navigate to `View collected data` for a survey that you've run
* In the "Analysis" section, try performing all of the actions that are listed as possible above 

Closes #95.


jeancochrane:
Requesting preliminary review from @beamalsky to get feedback on design and functionality.

jeancochrane:
Thanks for taking a look @beamalsky! I fixed the bug that was preventing reordering charts before saving them. Also made the charts full-width. I'll let you know once I have tests up.

beamalsky:
Those fixes worked, nice! Maybe for another issue but: it would be good to have "Move up" and "Move down" buttons/arrows in addition to the dragging interface. The chart areas are large enough that dragging more than one spot down is difficult on my laptop.

jeancochrane:
Great point @beamalsky. I'll add those buttons as part of this PR.

tests/test_dad.py
```diff
@@ -0,0 +1,138 @@
...
+    # Arguments:
+    #   * initial (int): Initial number of forms on the page at load time
+    #   * total (int): Total number of forms on the page at POST time
+    #   * forms (iterable): Iterable of dictionaries, each one representing the
+    #                       field value for a form in the formset
+    output_form = {
+        # Initialize with management form data
+        'form-INITIAL_FORMS': initial,
+        'form-TOTAL_FORMS': total,
+    }
```
jeancochrane: Including both the `initial` and `total` arguments in this function might seem a little bit superfluous, particularly because in these tests we only ever set `total` to `len(forms)`. In that case, why not just set `total = initial + len(forms)`? In the end I decided to keep `total` as an explicit argument because I wanted to preserve the possibility of distinguishing between creating and editing charts, as in the following cases:

1. Test creating new charts -> `total == initial + len(forms)`
2. Test editing existing charts -> `total == initial`

surveys/templates/survey_submitted_detail.html
```diff
@@ -30,9 +46,45 @@ <h2 class="card-title">{{ form_entry.name }}</h2>
...
+      <div id="sortable-chart-container">
+        {% for form in chart_formset %}
+          {% include 'partials/chart_form.html' with chart_id=forloop.counter0 %}
+        {% endfor %}
+      </div>
+    </form>
+
+    <button id="add-new-chart" class="btn btn-primary">Add new chart</button>
+
+    <!-- Use a hidden empty form as the base for adding new forms -->
```
beamalsky: Clever!

surveys/templates/survey_submitted_detail.html
```diff
@@ -44,6 +96,72 @@ <h2 class="card-title">{{ form_entry.name }}</h2>
       "leftColumns": 1
     }
   });
+
+  var $totalForms = $('[name$=TOTAL_FORMS]');
+  var chartContainer = '[class^="chart-container-"]'
+
+  function resetChartOrder() {
```
beamalsky: I'm not sure how difficult this is, but a moving chart animation would be really nice

tests/test_dad.py
beamalsky: Makes sense! Thanks for the well documented function

surveys/templates/partials/chart_form.html
```diff
@@ -0,0 +1,71 @@
...
+            </td>
+          </tr>
+        </tbody>
+      </table>
+      {% if form.errors %}
+        <div class="alert alert-danger mt-3">
+          <h4>Errors</h4>
+          <ul>
+            {% for field, errors in form.errors.items %}
+              <li>{{field}}: {{errors}}</li>
```
beamalsky: Could we use the field label here? So instead of `short description` in the screenshot below it would say "Short Description"
Screen Shot 2019-05-06 at 11 50 26 AM

surveys/templates/survey_submitted_detail.html
jeancochrane: Very true. A cursory Google search didn't return an immediately obvious way to do this, so I punted to #125.

surveys/templates/survey_submitted_detail.html
```diff
@@ -112,6 +112,18 @@ <h3 class="card-title">Analyze</h3>
     });
   }
 
+  function toggleMoveButtonVisibility() {
+    // Check to see if there is only one chart on the page, and if there is, hide
+    // the move buttons on that chart
+    if ($(chartContainer).filter(':visible').length === 1) {
```
beamalsky: Could we also check here if a single chart's position is first or last and hide the up and down arrows, respectively?

jeancochrane: Great point! Added.

Issue #90, opened 2019-04-17
Templated surveys


jeancochrane:
Our mockups suggest that we'll offer users the choice between creating a new survey from scratch and creating one from a "template". I don't think this is a required feature for the MVP, but I think it could be potentially very useful and I want to make sure that we keep track of it.

beamalsky:
We're doing this! 

When creating a new survey, users should be able to load a template with preset questions.

We'll wait on UCD to create these sample surveys on the staging site and copy them into templates. There may be different templates for Intercept and Observational surveys.

PR #121, opened 2019-05-01
Handle required PLDP metadata

### LINKED ISSUE #88: Pick a method for handling PLDP required fields
According to the PLDP, each survey needs a `Total` value attached to each row. `time_start`, `time_stop`, `representation`, and `method` are also required. 

I think we want to keep `total` required, but do we want to keep the other ones or make them optional? Should they have default values?

For observational surveys, it should not be possible to submit a survey without a value for `Total`. For intercept, that value should default to 1. (Think this though, and maybe ask UCD: is there any reason it wouldn't always be 1?)

I've created a `Total` form element in #87 that is required once it's added, but it still has to be added manually and can be deleted once it's there. Change that!

beamalsky:
## Overview

Following the discussion in #88, this PR removes `Total` as a survey question and makes `time_start`, `time_stop`, `representation`, and `method` questions automatically required fields once they are added to a survey. 

If we want to always create observational surveys with these fields, that should be handled once templated surveys (#90) have been implemented. Until we've figured out the templating and talked to UCD, I don't see a reason to spend more time making these elements required on every Observational survey.
 
## Testing Instructions

* Add `time_start`, `time_stop`, `representation`, and `method` questions. Are they automatically created as required questions?

 Closes #88

### DIFF
--- a/fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
+++ b/fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
@@ -45,11 +45,8 @@ class CollectDataPlugin(FormHandlerPlugin):
 
             new_survey = Survey.objects.create(**new_survey_info)
 
-            # total = self.get_saved_data('total', 1)
-
             new_survey_row = SurveyRow.objects.create(
                 survey=new_survey,
-                # total=total
             )
 
             meta_element_names = [name for (name, default) in meta_elements]
--- a/fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
+++ b/fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
@@ -45,11 +45,11 @@ class CollectDataPlugin(FormHandlerPlugin):
 
             new_survey = Survey.objects.create(**new_survey_info)
 
-            total = self.get_saved_data('total', 1)
+            # total = self.get_saved_data('total', 1)
 
             new_survey_row = SurveyRow.objects.create(
                 survey=new_survey,
-                total=total
+                # total=total
             )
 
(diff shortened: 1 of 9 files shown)


fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
```diff
@@ -45,11 +45,11 @@ def run(self, form_entry, request, form, form_element_entries=None):
 
             new_survey = Survey.objects.create(**new_survey_info)
 
-            total = self.get_saved_data('total', 1)
+            # total = self.get_saved_data('total', 1)
 
             new_survey_row = SurveyRow.objects.create(
                 survey=new_survey,
-                total=total
```
jeancochrane: I'd be in favor of removing these lines entirely instead of commenting them out. We can always come back to this PR if we want to reintroduce the functionality.

justspaces/settings.py
```diff
@@ -63,7 +63,7 @@
     'fobi.contrib.plugins.form_elements.fields.time',
 
     # custom metadata form elements
-    'fobi_custom.plugins.form_elements.fields.metadata.total',
+    # 'fobi_custom.plugins.form_elements.fields.metadata.total',
```
jeancochrane: Same here.

fobi_custom/plugins/form_handlers/collect_data/fobi_form_handlers.py
jeancochrane: Sorry, this should be attached to the new lines which were commented out instead of the old deleted lines.

beamalsky: Makes sense!

PR #122, opened 2019-05-01
Redirect to survey edit when survey is created

### LINKED ISSUE #114: Redirect to "Edit Survey" view after new survey creation

beamalsky:
## Overview

Redirect to "Edit Survey" detail view after new survey creation. Closes #114 

## Testing Instructions

* Create a new survey
* Are you redirected to the edit view for the survey you just created?

### DIFF
--- a/surveys/views.py
+++ b/surveys/views.py
@@ -2,6 +2,7 @@ from django.views.generic import TemplateView, ListView
 from django.views.generic.edit import CreateView, FormView
 from django.shortcuts import render, redirect
 from django.http import HttpResponseRedirect
+from django.urls import reverse_lazy
 
 from pldp.models import Agency, Location, Study, Survey
 
@@ -39,7 +40,6 @@ class SurveyCreate(CreateView):
     form_class = SurveyCreateForm
     model = SurveyFormEntry
     template_name = "survey_create.html"
-    success_url = '/'
 
     def form_valid(self, form):
         self.object = form.save()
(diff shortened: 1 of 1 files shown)


surveys/views.py
```diff
@@ -52,6 +52,12 @@ def form_valid(self, form):
 
         return HttpResponseRedirect(self.get_success_url())
 
+    def get_success_url(self):
+        form_entry_id = self.object.formentry_ptr_id
+        success_url = reverse_lazy('fobi.edit_form_entry', kwargs={'form_entry_id': form_entry_id})
```
jeancochrane: Any particular reason to use `reverse_lazy` instead of `reverse` here? Seems like both work, but I can't quite tell if there's an important difference in this context.

beamalsky: It would work either way here! Do you know an advantage to using plain `reverse`?

At first I tried defining `success_url` at the top of the class outside a method with `reverse`, and ran into [this issue](https://stackoverflow.com/questions/31275574/reverse-for-success-url-on-django-class-based-view-complain-about-circular-impor) which was solved by using `reverse_lazy`. When I switched to modifying `get_success_url` I didn't see a reason to change it back.

jeancochrane: Gotcha! No advantage I can think of eithe rway, I was just curious if there was a reason that the URLconf wouldn't be loaded at this point (as is the problem with in `get_success_url` that requires using `reverse_lazy` instead of `reverse`). I could see an argument for using `reverse` since it won't mislead another developer into thinking that there's a particular importance to `reverse_lazy` here, but both should work fine.

PR #123, opened 2019-05-03
Remove passphrase from server GPG key to fix deployments

### LINKED ISSUE #112: Remove passphrase from GPG key on the staging server
The staging server is failing to deploy because the GPG key has a passphrase. Either A) remove the passphrase from the key, or B) create a new key without a passphrase and update the Blackbox config to use the new key.

jeancochrane:
## Overview

Closes #112.

Remove the passphrase from the server GPG key in order to ensure that CodeDeploy can automatically decrypt application secrets.

## Notes

It had been so long since CodeDeploy ran successfully that the staging server database was incompatible with our migrations. I couldn't figure out how to restore properly from the dump in this repo (the dump was generated on Postgres 10 while the staging server runs 9.6, AFAIK that means they're not compatible) so in order to get CodeDeploy passing again I skipped the restore step and just created an empty database. If you've successfully restored that dump to the staging server, I'd love a quick tutorial on how to do it.

## Testing Instructions

 * Confirm that the manually-triggered CodeDeploy build for this PR ran successfully: https://console.aws.amazon.com/codesuite/codedeploy/deployments/d-HNRJEC4PZ?region=us-east-1


beamalsky:
Nice! In terms of db restore, I was also unable to get it working but I'm curious if we were running into different problems. I tried it following this gist: https://gist.github.com/beamalsky/417cc64989a1591aaf20880996d0a631

...and the restoration went smoothly, but I got a 500 server error when I tried to log in to the staging site. When I tried to drop and recreate an empty database (`createdb -U postgres -O datamade just-spaces`) I got the same server error. To get it working I eventually just triggered a new manual deployment, which creates an empty database if `just-spaces` doesn't exist. Could you tell me more about how your pg_restore was failing?

PR #145, opened 2019-05-23
Add API endpoint to return ACS data for DAD charts

### LINKED ISSUE #100: Add widget to DAD allowing user to compare to ACS data
Update the Data Analysis Designer to add a "Compare to" widget that can display ACS data. The "Compare to" widget should:

- Only display if the selected datasource has matching ACS data (see the spec: https://github.com/datamade/just-spaces/issues/78#issuecomment-483411198), and otherwise be hidden
- Show data for the correct datasource matching the survey question
- Allow the user to select a preset geographic resolution:
    - University City
    - Philadelphia
    - Pennsylvania
    - National
- Make the widget a multiselect in order to allow the user to choose multiple geographic resolutions to compare to
- Select colors for the ACS data bars randomly from a data viz palette

Depends on #99.

jeancochrane:
## Overview

Connects #100.

Update data imports for ACS data and add an endpoint that will return appropriate ACS data for a given survey question in the DAD.

### Tests and migrations

- Confirm that new tests pass on Travis
- Run migrations
- Run the new `import_data` command and confirm that it loads data:

```
python manage.py import_data
```

### Display of census geography widget and retrieval of ACS data

- Navigate to a survey submitted detail that has a question that is ACS-elegible (check `fobi_custom.plugins.form_elements.fields.types.TYPES_TO_ACS_VARIABLES` for a list of valid question types)
- Create a new chart for the eligible question and confirm that it displays a form field for Census geographies
- Open your browser's dev console, select a geographic level, and confirm that it prints ACS data to the console

### Proper hiding of census geography widget

- On the same page, change the `Primary data source` to the null value (`-----`) and confirm that the page hides the Census geographies field
- Navigate to a survey submitted detail that has a question that is **not** ACS-eligible
- Create a new chart for the ineligible question and confirm that the page does not show a Census geographies field


data/Makefile
```diff
@@ -1,28 +1,29 @@
...
-	 final/acs/means_of_transit.csv \
-	 final/acs/household_type.csv \
-	 final/acs/total_population.csv \
-	 final/acs/hours_worked_per_week.csv \
-	 final/acs/race.csv \
-	 final/acs/hispanic_or_latino_origin.csv \
-	 final/acs/household_tenure.csv \
-	 final/acs/vacancy_status.csv \
-	 final/acs/disability.csv
+all: final/acs/census_observations.csv
```
jeancochrane: Since an ACS count for a given FIPS code and variable is now stored as a `CensusObservation`, I updated the import script to merge all of the ACS data together. This is intended to follow our new best practices for data imports, where as much processing as possible is done in the Makefile and the `import_data` management command is scoped to simply loading the data into the database.

data/scripts/merge_acs_data.py
```diff
@@ -0,0 +1,36 @@
...
+
+if __name__ == '__main__':
+    # Parse input files from the command line
+    filenames = sys.argv[1:]
+    if len(filenames) == 0:
+        raise NameError(
+            'One or more arguments representing the files to transform are required'
+        )
+
+    fieldnames = ['fips', 'variable', 'fields']
```
jeancochrane: You can get a better sense of what each of these fields is supposed to represent in the `CensusObservation` model.

fobi_custom/plugins/form_elements/fields/types.py
```diff
@@ -45,3 +45,17 @@
...
+    'gender_observational': 'age_and_sex',
+    'age_intercept': 'age_and_sex',
+    'gender_intercept': 'age_and_sex',
+    'income': 'household_income',
+    'education': 'educational_attainment',
+    'race': 'race',
+    'employment': 'employment_status',
+    'household_tenure': 'household_tenure',
+    'transportation': 'means_of_transit'
+}
```
jeancochrane: This map helps us determine which types of survey questions can be associated with ACS data in the DAD.

It's possible we could store this information somewhere on a model, perhaps `SurveyComponent` or somewhere in `django-pldp`? I just couldn't really figure out what the right level of abstraction was.

surveys/management/commands/import_data.py
```diff
@@ -5,53 +5,54 @@
...
 
+from surveys import models
+
 DB_CONN = 'postgresql://{USER}:{PASSWORD}@{HOST}:{PORT}/{NAME}'
 DB_CONN_STR = DB_CONN.format(**settings.DATABASES['default'])
 
+CENSUS_AREAS = {
+    'USA': ['1'],
+    'Pennsylvania': ['42'],
+    'Philadelphia': ['42101']
```
jeancochrane: The FIPS codes are stored in lists so that when we create new areas we can have them be groups of Census tracts (e.g. University City will be a list of tracts). I started simple here.

surveys/models.py
```diff
@@ -36,6 +40,48 @@ class SurveyFormEntry(FormEntry):
...
+        return self.name
+
+    def get_observations_from_component(self, component_name):
+        """
+        Given the name of a SurveyComponent, return the ACS observation data for this area.
+
+        :param component_name: The name of a SurveyComponent
+        :returns: A dictionary representing the stored data of the CensusObservation
+                  objects for the given area and SurveyComponent
+        """
```
jeancochrane: At first I tried directly associating `CensusObservation`s to `CensusArea`s by having a `ManyToManyField` called `CensusArea.observations`. This created a problem because a `CensusArea` is in a sense more abstract than a `CensusObservation`; it's a collection of FIPS codes, not a collection of specific observations.

This method is intended to help bridge the gap by returning `CensusObservations` for a given `CensusArea` based on the `CensusArea.fips_code` attribute and `SurveyComponent.name` ID. The use of `SurveyComponent.name` is a bit tricky, but makes sense for our use-case: The clearest way to define this relationship would be to use an ACS variable name instead (like `age_and_sex` or `household_tenure`), but since the `SurveyComponent.name` attribute is available to us directly from the survey submitted data object, it's a lot easier to use it to find `CensusObservation`s here.

surveys/urls.py
```diff
@@ -6,53 +6,51 @@
...
 """
 from django.conf.urls import url, include
+from django.urls import path
 from django.utils.translation import ugettext_lazy as _
 from django.contrib.auth.decorators import login_required
-from surveys.views import AgencyCreate, LocationCreate, StudyCreate, \
-                            StudyAreaCreate, SurveyCreate, SurveyPublish, \
-                            SurveyListEdit, SurveyListRun, Signup, SurveySubmittedList, \
-                            SurveySubmittedDetail
+from surveys import views
```
jeancochrane: Rather than add another import to this long list, I did a quick refactor to use the module name to reference the view methods.

fobi_custom/plugins/form_elements/fields/types.py
beamalsky: This makes sense to me!

surveys/management/commands/import_data.py
beamalsky: Nice!

surveys/urls.py
beamalsky: Thanks! I'd been thinking about doing this.

Issue #140, opened 2019-05-21
Automatically select user on survey creation


beamalsky:
The top field here should grab who's logged in and autofill:
<img width="864" alt="Screen Shot 2019-05-21 at 3 54 37 PM" src="https://user-images.githubusercontent.com/1094243/58130093-d6e91100-7be0-11e9-9716-5a89d8fc50db.png">

I can't think of a case where the current user isn't the one who should be saved.

jeancochrane:
Perhaps we could hide the `User` field when we prepopulate it? I think you're right that there's no case where the current user isn't the right one, but also I don't think we want to allow a user to claim that another user created a survey.

beamalsky:
I think you're right!

PR #147, opened 2019-05-29
Issue 106

### LINKED ISSUE #106: Expand Survey Properties editing tab
Screen Shot 2019-04-22 at 1 57 11 PM

This ^^ 

We'll also want to expand the form under "Properties" to include our new expanded Survey metadata.

beamalsky:
## Overview

Closes #106. This PR creates a view for editing survey properties. It uses the same form as the `SurveyCreate` view, and user will be automatically selected and hidden in #140. 

## Testing Instructions

* Create a new survey
* The site should take you to the "Edit Survey" page. Can you find where to change its properties? When you save them, does it bring you back to the "Questions" tab?

### DIFF
--- a/surveys/views.py
+++ b/surveys/views.py
@@ -1,6 +1,6 @@
 import json
 
-from django.views.generic import TemplateView, ListView
+from django.views.generic import TemplateView, ListView, UpdateView
 from django.views.generic.edit import CreateView, FormView
 from django.shortcuts import render, redirect
 from django.http import HttpResponseRedirect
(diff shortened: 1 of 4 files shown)


surveys/views.py
```diff
@@ -138,6 +138,19 @@ def get_success_url(self):
...
+    model = SurveyFormEntry
+    template_name = "survey_properties_edit.html"
+    form_class = SurveyCreateForm
+    context_object_name = 'form_object'
+
+    def get_success_url(self):
+        form_entry_id = self.object.pk
+        success_url = reverse_lazy('fobi.edit_form_entry', kwargs={'form_entry_id': form_entry_id})
+
+        return str(success_url)
```
jeancochrane: What's the reason why the URL needs to be cast to `str`?

beamalsky: It's unnecessary, good catch! I think just copied over from an example without checking.

PR #149, opened 2019-06-03
Add basic CSV export

### LINKED ISSUE #80: Export collected data as CSV
Update the collected data table to allow the user to:

- sort and filter columns
- export data displayed in the collected data table as CSV

beamalsky:
## Overview

Closes #80. Implements basic DataTables CSV export button and improves DataTables styling. Intended to look like this:

Screen Shot 2019-06-03 at 4 30 28 PM

## Testing Instructions

 * Look at a survey with submitted data
* Can you view that data in the table on the page?
* Click the "Download as CSV" button. Does it look like you'd expect?


surveys/templates/survey_submitted_detail.html
```diff
@@ -2,11 +2,18 @@
...
 {% endblock %}
 
 {% block extrajs %}
-  <script type="text/javascript" src="{% static 'js/datatables.min.js' %}"></script>
+  <script type="text/javascript" src="{% static 'js/jquery.dataTables.min.js' %}"></script>
+  <script type="text/javascript" src="{% static 'js/dataTables.bootstrap4.min.js' %}"></script>
+
+  <script type="text/javascript" src="{% static 'js/dataTables.buttons.min.js' %}"></script>
+  <script type="text/javascript" src="{% static 'js/buttons.bootstrap4.min.js' %}"></script>
+  <script type="text/javascript" src="{% static 'js/buttons.html5.min.js' %}"></script>
```
jeancochrane: Could you say a little bit about what these file changes are accomplishing?

surveys/templates/survey_submitted_detail.html
```diff
@@ -97,16 +104,14 @@ <h3 class="card-title">Analyze</h3>
...
-  $('#survey-submitted-detail').DataTable({
-    // scrollx styling taken from
-    // https://stackoverflow.com/questions/45136218/datatables-scrollx-causing-squashed-header
-    "sScrollX": "100%",
-    "sScrollXInner": "110%",
-    "fixedColumns":   {
-      "leftColumns": 1
-    }
+  var table = $('#survey-submitted-detail').DataTable({
+    buttons: [{ extend: 'csv', text: 'Download as CSV' }],
```
jeancochrane: Because I'm a sucker for icons, how would you feel about a download icon here?

```suggestion
    buttons: [{ extend: 'csv', text: '<i class="fa fa-fw fa-download mr-1"></i>Download as CSV' }],
```

Screenshot from 2019-06-03 16-47-49

surveys/templates/survey_submitted_detail.html
```diff
@@ -97,16 +104,14 @@ <h3 class="card-title">Analyze</h3>
...
+    lengthChange: false,
   });
 
+  table.buttons().container()
+      .appendTo( '#survey-submitted-detail_wrapper .col-md-6:eq(0)' );
```
jeancochrane: Why do we need the additional `.col-md-6:eq(0)` selector here? Shouldn't the ID selector uniquely identify the element?

surveys/templates/survey_submitted_detail.html
jeancochrane: It'd also be nice to use the `filename` attribute to set the CSV filename based on the title of the survey.

beamalsky:
I'm going to make these changes as a separate PR since I accidentally merged #150 into `master` without changing the base branch!

surveys/templates/survey_submitted_detail.html
beamalsky: Following the DataTables [documentation](https://datatables.net/extensions/buttons/examples/initialisation/export.html):

> Buttons provides button types that will automatically determine if HTML5 or Flash should be used based on the browser's functionality and it is strongly recommended that you use these button types over the specific HTML5 or Flash button types. These are: copy, csv, excel, pdf.

I'd forgotten the flash js file and I've added it here!

surveys/templates/survey_submitted_detail.html
beamalsky: Without `.col-md-6:eq(0)` this button would be appended to the end of the table, when instead I want it to be the first element in the row containing the search box. 

`.col-md-6:eq(0)` tells it to append the button once after an instance of a `.col-md-6` element

Reference: https://datatables.net/extensions/buttons/examples/styling/bootstrap4.html

beamalsky:
Here is that PR: https://github.com/datamade/just-spaces/pull/152

Sorry to make this more complicated!

PR #153, opened 2019-06-04
Automatically set user when creating new survey forms

### LINKED ISSUE #140: Automatically select user on survey creation

beamalsky:
## Overview

Closes #140. This PR automatically sets the user in the survey form creation form and hides that field. 

## Testing Instructions

* Create a new survey (`surveys/create/`) and make sure there's no user field displayed
* Make sure tests pass. If so, this works! But you can dig deeper: 
* Go into the psql database. Note the `id` of the user you're logged in as (`select * from users_justspacesuser`)
* Create a new survey
* Look at it in the database (`select * from fobi_formentry;`). Does it have the correct user assigned?

### DIFF
--- a/surveys/views.py
+++ b/surveys/views.py
@@ -8,8 +8,7 @@ from django.urls import reverse_lazy
 from django.forms import modelformset_factory
 from django.contrib import messages
 
-from pldp.models import Agency, Location, LocationArea, LocationLine, Study, \
-                    StudyArea, Survey
+from pldp import models as pldp_models
 
 from users.models import JustSpacesUser
 from users.admin import JustSpacesUserCreationForm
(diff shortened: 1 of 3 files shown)


surveys/views.py
```diff
@@ -8,32 +8,30 @@
...
 from users.models import JustSpacesUser
 from users.admin import JustSpacesUserCreationForm
 
 from fobi.views import add_form_handler_entry
 
 from .models import SurveyFormEntry, SurveyChart
-from .forms import StudyCreateForm, StudyAreaCreateForm, SurveyCreateForm, \
-                   SurveyChartForm, LocationCreateForm, LocationAreaCreateForm, \
-                   LocationLineCreateForm, AgencyCreateForm
+from surveys import forms as survey_forms
```
jeancochrane: Good refactor here!

Issue #84, opened 2019-04-15
Deleting surveys


beamalsky:
While working on #83 I've been thinking about how we want to handle deleting surveys. Currently, the Survey list view has a button to delete SurveyFormEntries. 

I don't think users should be able to do that once a survey is published, since you should be able to look back and rerun a survey once you've started collecting data on it! Maybe we need some kind of additional "active" vs. "inactive" state—for example, if a UCD admin wants to mark when data collection for a survey has concluded—or to let users set a start and (optional) end date on surveys when they're creating them, then only show surveys that are within their active date ranges.

This is related to #54. This is a temp solution I just made for handling deleted surveys we've already collected data for: https://github.com/datamade/just-spaces/pull/83/commits/8e2671fda7f008b82266d2abb5729af5d88a619e

jeancochrane:
If we decide to support deletion, I wonder if we could [add a custom constraint](https://www.fusionbox.com/blog/detail/custom-database-constraints-in-django/594/) to the Survey table with a migration in the `surveys` app? The constraint I'm imagining would be something like:

```sql
ALTER TABLE pdlp_survey
ADD CONSTRAINT form FOREIGN KEY (form_id)
REFERENCES surveys_surveyform (id) ON DELETE CASCADE
```

I'm not really sure if this would work since I don't know if we can reliably say that the `pldp_survey` table will exist before the `surveys_surveyform` table in the app (I would guess that the order of the apps in `INSTALLED_APPS` would determine this). Still, I think it's worth a shot.

beamalsky:
Proposal: We add a new attribute to the SurveyForm model: `is_active`. On new surveys this defaults to true, and once a survey is published it can be marked `active` or`inactive` from the list view and filtered by that value. 

In addition, the Edit Survey list view will include a delete button to fully delete surveys. As these have not been published, they will not have any saved data and can be safely deleted.

jeancochrane:
Sounds good to me!

PR #154, opened 2019-06-04
List, detail, and delete views for studies and locations

### LINKED ISSUE #134: Round out location and study views
We'll want simple ways for users to interact with locations and studies. These should include: 

- Location list
- Location view
- Location edit?
- Study list
- Study view
- Study edit

beamalsky:
## Overview

Closes #134. This PR creates list, edit, and delete views for PLDP Study and Location models. This means that users can look at details of studies and locations once they're created, and that we should make sure that only some users have permission to do the deleting once we finalize #48.

## Testing Instructions

* Run migrations
* Navigate to `http://127.0.0.1:8000/surveys/create/`
* In the `Location` and `Study` field you should see links to list and detail views for each. Click around and make sure all behave as you'd expect, and that you can successfully deactivate both once created.


beamalsky:
@jeancochrane this isn't yet ready for review. I realized that deleting a Location or Study right now deletes any Survey that uses it, which isn't what we want! This is related to #84 and will take a little more thinking.

jeancochrane:
Cool, thanks for letting me know @beamalsky! I'll hold off on taking a look until you re-request.

beamalsky:
Plan after chatting with Jean: I'm going to add `is_active` fields to the PLDP Study and Location models, then filter out the inactive ones in these list views. I'll also create edit views for both in #155

beamalsky:
@jeancochrane this is now ready for your eyes! It has some merge conflicts with master, which have to do with renaming the imports and should be a quick fix once this is ready to go.

surveys/static/js/location_create.js
```diff
@@ -11,15 +11,17 @@ typeDropdown.change(function() {
...
   var $locationLine = $('*[id^="div_id_location-line-"]');
 
   var $locationLineDateMeasured = $('#id_location-line-date_measured');
   var $locationAreaDateMeasured = $('#id_location-area-date_measured');
 
-  var selectedVal = typeDropdown.value;
+  var selectedVal = typeDropdown[0].value;
 
   if (selectedVal == 'area') {
+    console.log("area selected!");
```
jeancochrane: Looks like a couple stray print statements in this file.

surveys/templates/location_delete.html
```diff
@@ -0,0 +1,15 @@
...
+{% block content %}
+<div class="col col-lg-10 card">
+  <div class="card-body col-lg-9">
+    <h2 class="card-title">Delete Location</h2>
+
+     <p class="card-text">Are you sure you want to delete <b>{{object}}</b>?</p>
+     <p class="card-text">Once locations are deleted they cannot be recovered.</p>
+
+     <form method="POST" class="post-form">{% csrf_token %}
+     <button type="submit" class="save btn btn-primary">Delete</button>
```
jeancochrane: Could we have a `Cancel` button here, similar to the other pages?

surveys/templates/location_detail.html
```diff
@@ -0,0 +1,57 @@
+{% extends 'base.html' %}
+{% load static %}
+{% load leaflet_tags %}
+{% load crispy_forms_tags %}
+
+{% block extracss %}
+  {% leaflet_css %}
```
jeancochrane: I'm not totally sure why, but the map loads at a height on 300px on my machine, which looks a little bit squished:

Screenshot from 2019-06-05 14-21-44

surveys/templates/location_detail.html
```diff
@@ -0,0 +1,57 @@
...
+          <td style="width: 30%"><b>{{label}}</b></td>
+          <td>{{content}}</td>
+        </tr>
+        {% endif %}
+      {% endfor %}
+
+    </tbody>
+    </table>
+
+  <a class="btn btn-secondary float-right" href="{% url 'locations-delete' location.id %}">Delete study</a>
```
jeancochrane: Small typo here:

```suggestion
  <a class="btn btn-secondary float-right" href="{% url 'locations-delete' location.id %}">Delete location</a>
```

surveys/templates/location_list.html
```diff
@@ -0,0 +1,47 @@
...
+    <br />
+
+    <table id="locations-list" class="table table-hover table-striped table-responsive-sm card-text">
+    <thead>
+      <tr>
+        <th scope="col">Primary Name</th>
+        <th scope="col">Geometry Type</th>
+        <th scope="col">Subdivision</th>
+        <th scope="col">Character</th>
+        <th scope="col">Delete</th>
```
jeancochrane: Since we're centering the Delete icon, we should probably center this column name as well:

```suggestion
        <th class="text-center" scope="col">Delete</th>
```

surveys/templates/study_delete.html
```diff
@@ -0,0 +1,15 @@
...
+{% block content %}
+<div class="col col-lg-10 card">
+  <div class="card-body col-lg-9">
+    <h2 class="card-title">Delete Study</h2>
+
+     <p class="card-text">Are you sure you want to delete <b>{{object}}</b>?</p>
+     <p class="card-text">Once studies are deleted they cannot be recovered.</p>
+
+     <form method="POST" class="post-form">{% csrf_token %}
+     <button type="submit" class="save btn btn-primary">Delete</button>
```
jeancochrane: A cancel button would be nice here, too.

surveys/templates/study_list.html
```diff
@@ -0,0 +1,47 @@
...
+    <br />
+
+    <table id="studies-list" class="table table-hover table-striped table-responsive-sm card-text">
+    <thead>
+      <tr>
+        <th scope="col">Title</th>
+        <th scope="col">Agency</th>
+        <th scope="col">Manager</th>
+        <th scope="col">Manager Email</th>
+        <th scope="col">Delete</th>
```
jeancochrane: Same comment here RE: centering the Delete button header.

surveys/urls.py
```diff
@@ -8,51 +8,73 @@
...
-        login_required(AgencyCreate.as_view()),
+        login_required(survey_views.AgencyCreate.as_view()),
         name='agencies-create'),
 
     url(r'locations/create/$',
-        login_required(LocationCreate.as_view()),
+        login_required(survey_views.LocationCreate.as_view()),
         name='locations-create'),
 
+    url(r'locations/list/$',
```
jeancochrane: This is more of a style comment, but I wanted to call out here that our URLpatterns have started to diverge from the [REST guidelines](https://medium.com/hashmapinc/rest-good-practices-for-api-design-881439796dc9) that we were working with earlier. In particular:

> 3. Let the HTTP Verb Define Action

Following this advice, I think `/entities/list/` and `/entities/detail/<pk>/` put too much of the verb in the URL, and I wonder if they could be replaced by `/entities/` and `/entities/<pk>/` respectively. (In this case, you could swap `locations` or `studies` for `entities`.)

There's some amount of necessary divergence we need to take from this guideline, e.g. since we're using a `DeleteView` for `delete` actions that accepts both `GET` and `POST` requests it makes a lot of sense to have `/entities/delete/` routes. But for `list` and `detail` routes I wonder whether we could get away with simpler URLs.

surveys/views.py
```diff
@@ -91,6 +93,62 @@ def post(self, request, **kwargs):
...
+    template_name = "location_detail.html"
+    context_object_name = 'location'
+
+    def get_context_data(self, **kwargs):
+        context = super(LocationDetail, self).get_context_data(**kwargs)
+
+        location = context['location']
+
+        location_area = get_or_none(pldp_models.LocationArea, location=location)
+        location_line = get_or_none(pldp_models.LocationLine, location=location)
```
jeancochrane: Do we need this custom function (`surveys.utils.get_or_none`)? If we used `try`/`except`/`else` blocks for this conditional flow instead, we could avoid introducing a new module with a specialty function.

tests/test_views.py
```diff
@@ -23,6 +23,33 @@ def test_study_create(client, user, study_area):
...
+    response = client.get(url)
+
+    assert response.status_code == 200
+
+
+@pytest.mark.django_db
+def test_study_delete(client, user, study):
+    client.force_login(user)
+    url = reverse('studies-delete', kwargs={'pk': study.id})
+    response = client.get(url)
```
jeancochrane: I think this only tests that we can `GET` the delete page, but not that we can actually delete a study, right?

tests/test_views.py
```diff
@@ -23,6 +23,33 @@ def test_study_create(client, user, study_area):
     assert study_area.name in response.content.decode('utf-8')
 
 
+@pytest.mark.django_db
+def test_study_list(client, user):
+    client.force_login(user)
+    url = reverse('studies-list')
+    response = client.get(url)
+
+    assert response.status_code == 200
```
jeancochrane: Do we have any location or study fixtures that we could require for these `list` tests to make sure that all locations/studies are getting displayed in the list?

tests/test_views.py
```diff
@@ -23,6 +23,33 @@ def test_study_create(client, user, study_area):
...
+
+
+@pytest.mark.django_db
+def test_study_detail(client, user, study):
+    client.force_login(user)
+    url = reverse('studies-detail', kwargs={'pk': study.id})
+    response = client.get(url)
+
+    assert response.status_code == 200
```
jeancochrane: Could we test to see that all the relevant information is displayed in the table?

surveys/static/js/location_create.js
beamalsky: Oops!

surveys/templates/location_detail.html
beamalsky: This was the intended height, but now that you mention it does look a bit small. I'll make it longer!

surveys/templates/location_delete.html
beamalsky: Good call

surveys/templates/location_detail.html
beamalsky: Got it!

surveys/templates/location_list.html
beamalsky: 👍🏻I'll get this for all the lists

surveys/urls.py
beamalsky: I like it!

surveys/views.py
beamalsky: I wrote this custom function to keep the code DRYer, as I'd be writing the same try/except block over and over. I've found this is an operation I've wanted in most Django projects, so it seems handy. Is there a reason I should write out those blocks repeatedly?

tests/test_views.py
beamalsky: Civic Exchange internet didn't push through my last commit extending this test 😬 It's in now testing both get and post!

tests/test_views.py
beamalsky: Yep! I'll make new fixtures for inactive Locations and Studies too.

tests/test_views.py
beamalsky: Yep!

surveys/views.py
jeancochrane: Gotcha. I don't necessarily agree that the `try`/`except` flow violates DRY, since it's really only a couple lines longer than the `get_or_none` solution:

```python
location_area = get_or_none(pldp_models.LocationArea, location=location)

if location_area:
    ...
```

vs.

```python
try:
    location_area = pldp_models.LocationArea.objects.get(location=location)
except pldp_models.LocationArea.DoesNotExist:
    pass
else:
    ...
```

But in the end this is a matter of preference, so I'm fine going with the method if you prefer it. Agreed that I'm surprised there isn't a built-in way of retrieving `None` for missing objects using the `get` method in the ORM.

PR #157, opened 2019-06-05
Display ACS data for selected questions in DAD

### LINKED ISSUE #100: Add widget to DAD allowing user to compare to ACS data

jeancochrane:
## Overview

Closes #100.

Update the Data Analysis Designer to allow the user to add ACS data to charts for specific survey questions.

## Testing Instructions

* Create a survey that has all preset Intercept questions
    * Add a few custom Intercept questions so that you can make sure that the DAD doesn't allow you to compare them to ACS data
* Create a survey that has the following preset Observational questions:
    *  Age, all detail levels
    * Gender
    * Add a few custom Observational questions as well
* Publish your surveys and run them a few times with dummy data
* Navigate to the submitted survey view for each survey and confirm:
    * You are shown ACS geography options for qualifying questions
    * You can select different combinations of ACS geographies and see the data displayed on the chart
    * You are not shown ACS geography options for the custom questions you created
* Save some charts with ACS data, and confirm that when you reload the page they load with the data intact


surveys/models.py
```diff
@@ -78,9 +84,7 @@ def get_observations_from_component(self, component_name):
                 )
         else:
             raise CensusObservation.DoesNotExist(
-                'No SurveyComponent object found with the name: {}'.format(
-                    component_name
-                )
+                'No SurveyComponent found with the name: {}'.format(component_name)
```
beamalsky: Looks like we'll need a new migration to reflect this. When I run `python manage.py migrate` I get this:

```
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

fobi_custom/plugins/form_elements/fields/intercept/household_tenure/fobi_form_elements.py
```diff
@@ -9,7 +9,7 @@ class HouseholdTenurePlugin(FormFieldPlugin):
     """HouseholdTenurePlugin."""
 
     uid = "household_tenure"
-    name = "How many years have you lived at your current address?"
+    name = "What year did you move into your current address?"
```
beamalsky: Should this be a dropdown field instead of an integer field?

fobi_custom/plugins/form_elements/fields/types.py
```diff
@@ -48,14 +43,39 @@
...
-    'age_observational': 'age_and_sex',
-    'gender_observational': 'age_and_sex',
-    'age_intercept': 'age_and_sex',
-    'gender_intercept': 'age_and_sex',
-    'income': 'household_income',
-    'education': 'educational_attainment',
-    'race': 'race',
-    'employment': 'employment_status',
-    'household_tenure': 'household_tenure',
-    'transportation': 'means_of_transit'
```
beamalsky: The prewritten question here is "How did you travel here?," which I'm guessing is not the same phrasing as the census question. What do you think about including a note with charts that include Census data to say where the information is coming from, including the framing of the question?

fobi_custom/plugins/form_elements/fields/intercept/household_tenure/fobi_form_elements.py
jeancochrane: I think it works fine as an integer field, but we could definitely change it to a dropdown where the user selects from a set of bins. To me it seems like the cost incurred by that approach would be that custom analyses of the exported data would have to be limited in resolution by our preset bins. Do you think there's an upside to making it a select field that would offset that cost?

fobi_custom/plugins/form_elements/fields/types.py
jeancochrane: Good point. We had also talked about adding a similar caveat for racial data, since ACS doesn't record Hispanic/Latino as a race.

Right now, the pull request assumes that it's the responsibility of the user to note these kinds of caveats in the "short description" field in the DAD. Adding custom notes seems like it would have the advantage of taking that step off of the user's plate, but at the cost that the notes would either have to 1) be fixed by us or 2) exposed for list/detail/edit through separate views. It would also mean that notes would have to be global to data sources, and couldn't be customized on a per-chart basis (whereas the short description field can).

What do you think about opening this as a separate issue and raising it with UCD? I'd be curious to hear if they'd rather have more flexibility as with the short description field, or if they'd rather save time with preset notes.

surveys/models.py
jeancochrane: I think the missing migration is actually due to the `SurveyChart.census_areas` attribute. Oops! I'll add it in.

fobi_custom/plugins/form_elements/fields/intercept/household_tenure/fobi_form_elements.py
beamalsky: I was actually imagining a dropdown with single values ranging from, say, 1919 - 2019. That would prevent people from entering values in the future or too far in the past to be possible. Now that I'm writing this out, though, I realize there's a better solution: setting a default minimum and maximum value on this field, like the one that's currently on the Number question. What do you think? Maybe this is just a nice-to-have.

jeancochrane: I like that idea a lot! I'll add it in. The minimum/maximum is enforced on the Fobi level, right? If so, then I can set the minimum to something like `1900` and set the maximum to the current year.

fobi_custom/plugins/form_elements/fields/types.py
beamalsky: Raising this with UCD sounds like the right route here to me too—maybe we can add it to our MVP demo agenda. 

Could you say more about #2 above when you open the issue? I don't quite understand what you mean. 

I don't think it makes sense to task the user with noting how the site is using ACS data, as they wouldn't necessarily know what's being pulled in or how it's being fit to the questions. For that reason, I think it makes sense to have ACS data notes as a global caption on a field wherever it's used, and treated as a separate attribute from the Short Description box, which I've been thinking of as more of a survey-specific caption.

beamalsky:
Go ahead and merge this once you've had a chance to look at my last comments!

fobi_custom/plugins/form_elements/fields/types.py
jeancochrane: Opened up #160 to track this and added an agenda item to our MVP call to make sure we discuss it! By "exposed for list/detail/edit through separate views" I mostly meant that if we want to allow the user to create, edit, or delete these kinds of notes, we'll need corresponding views for it. Not a huge deal, of course, it'll just require a couple extra hours of work and some thinking about how to integrate it with the existing menu.

Ultimately I think you're right, UCD will almost certainly want these attributes and they'll want to be able to edit them. But mostly I just want to hear them vocalize that before putting the work into it.

fobi_custom/plugins/form_elements/fields/intercept/household_tenure/fobi_form_elements.py
jeancochrane: I thought enforcing min/max for IntegerFields would be an out-of-the-box configuration in Fobi but it looks like it'll require looking through the source and copying their implementation, so I kicked this off to #161.

PR #162, opened 2019-06-10
Deleting/deactivating surveys

### LINKED ISSUE #84: Deleting surveys

beamalsky:
## Overview

Closes #84. In design review with Derek we decided to treat deactivation as a soft delete, and refer to it as such on the site. This PR adds buttons to delete (read: deactivate) both published and unpublished surveys. It also creates a `deactivate.html` partial to cut down on repetitive code on the templates for deleting locations, studies, and surveys.

## Testing Instructions

* Run migrations
* Create a new survey. Make sure you can see it in the "Edit Surveys" view
* Delete that survey. Make sure it disappears
* Create a new published survey or pick an existing one. 
* Delete it. Make sure that deactivated survey no longer shows up in the "Run survey" list


surveys/models.py
```diff
@@ -39,6 +38,11 @@ class SurveyFormEntry(FormEntry):
         default='intercept'
     )
 
+    active = models.BooleanField(default=True)
```
jeancochrane: This attribute looks good! I'll note that it's a common convention, in Python and elsewhere, to use predicate clauses to name boolean variables and attributes (e.g. `active` would be `is_active`). See this discussion on SE: https://softwareengineering.stackexchange.com/a/221873

This is a common convention related to style and it's not enshrined in any of our best-practices documents, so don't worry about changing your attribute name. I only wanted to point out the convention for your consideration.

Issue #165, opened 2019-06-10
Feedback from Derek ahead of our demo


jeancochrane:
### General

- [x] Create ACS geographies on staging site by running `python manage.py import_data`
- [x] Run `python manage.py initialize_pldp` on staging site
- [x] Bump the size of header-level help text
- [x] Rename “deactivate” to “delete,” with help text explaining that it’s a soft delete

### List views
- [x] Add explainer text for top-level entities (Studies, Surveys, Study Areas, Locations)
- [x] Add “Create new” button to all list view templates
- [ ] Combine list table buttons into “Actions” column on list views (moved to #168)

### Navbar
- [x] Change “Create new…” to “Create survey”
- [x] Move DataMade site credit to the footer

### Surveys
- [x] Figure out if Study and Location should be required fields; update the data handler appropriately (Update: they should be required!)

### Adding survey questions view
- [x] Display the question options when adding a preset question to a survey
- [x] Change “Label” -> “Question text”
- [x] Change “Default” -> “Default answer”
- [x] Add help text for each form element

### Survey edit view

- [ ] Add dragging icon to draggable survey questions. (Still to do: Make sure that the draggable cursor appears on the left side of the question form on Survey Edit; moved to #169)
- [x] Change “Save ordering” button to “Save and exit”
- [x] Add a “Preview” button
- [x] Make Survey.type not editable on the properties page

### Run Survey view

- [x] Make Run Survey icon into a button
- [x] Add to table metadata:
    - [x] Number of questions
    - [x] Last run
- [x] Add the name of the survey to “Survey submitted!” confirmation page

### Survey submitted list view
- [x] Show number of runs of the survey

### Survey submitted detail view
- [x] Add help text to Census geographies selectbox explaining that you can select multiple
- [x] Clean up styles of Analyze form (Select2, bootstrap inputs)
- [x] Change “Primary source” label to say “Survey question”

PR #167, opened 2019-06-11
Ensure that ACS data selections get saved

jeancochrane:
## Overview

In the course of addressing #165, @beamalsky noticed that ACS selections weren't getting saved in the DAD. This chart updates the route and frontend scripts to make sure that ACS data gets saved and displayed on first load.

## Testing Instructions

* Create a chart with an ACS-compatible question (like `How old are you?`)
* Select two ACS geographies
* Save the charts
* Confirm that when you reload the page, the chart displays the ACS data

### DIFF
--- a/surveys/templates/survey_submitted_detail.html
+++ b/surveys/templates/survey_submitted_detail.html
@@ -198,6 +198,37 @@
       }
     });
 
+    function displayAcsSeries(selectbox) {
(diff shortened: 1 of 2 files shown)


beamalsky:
It works! Thanks for jumping on this. Go ahead and merge if you're ready.

PR #172, opened 2019-06-18
Constrain content width on large screens

jeancochrane:
## Overview

Make sure that long content isn't too wide on large screens.

## Testing Instructions

* Visit a collected data view with lots of columns and confirm that the page doesn't require a horizontal scroll
* Use browser dev tools to preview the site on mobile and confirm that the width still looks good


beamalsky:
Looks great! Thanks for doing this last minute.

Issue #177, opened 2019-06-18
Create CRUD views for CensusAreas


jeancochrane:
Add Create, Read (i.e. List), Update (i.e. Edit), and (soft) Delete views for CensusArea objects. Similar to Locations, additionally update the "Compare to Census" question in the DAD prompting the user to create a geography if they don't see one they want.

For the scope of this issue, the Create and Edit views need only add/update the `CensusArea.name` attribute. (The geographies themselves should be added in #101.) The List view should show the name of the CensusArea in addition to a thumbnail map so that the user can quickly scan to get a rough sense of what boundaries the CensusArea are.

beamalsky:
I could also take this on if it makes sense!

Issue #101, opened 2019-04-19
Create geographic widget for selecting CensusAreas


jeancochrane:
Create a form widget allowing a user to select a group of census tracts or blocks to create a CensusArea. The widget should:

- Display all available Census blockgroups on a map
- Allow the user to select, deselect, and clear multiple blockgroups from the map
- Save the blockgroups to `CensusArea.fips_codes`

We built out some prior art for this kind of "geographic multiselect" in EDI. See: 
- https://github.com/datamade/erikson-edi/blob/ad68b7853b967be3e1adc3cf6330b8c5e63218fb/edi/forms.py#L59-L68
- https://github.com/datamade/erikson-edi/blob/ad68b7853b967be3e1adc3cf6330b8c5e63218fb/edi/widgets.py#L9-L53

beamalsky:
Would you be up for including help text on Census Areas and how to use them in this PR?

jeancochrane:
I think help text might make more sense in #177, since the help text will primarily be a part of the views and not the widget, right?

beamalsky:
We'll need help text describing how to use the widget as well as help text explaining how census areas are defined and how to use them, both of which I think you're better positioned to write.

jeancochrane:
Gotcha. I'm still not sure that text will make sense in the context of this PR, but I'm happy to take responsibility for making sure it gets incorporated once we wire the widget and the models together.

beamalsky:
Sounds good!

PR #181, opened 2019-06-19
Add CRUD views for CensusAreas

beamalsky:
## Overview

Connects #177. This PR creates Create, List, Edit, and Deactivate views for CensusAreas. For now, the Create and Edit views only display the `name` field, meaning new CensusAreas cannot actually be created. This PR also adds an `is_active` attribute to CensusAreas in order to use them with our deactivate method.

## Testing Instructions

 * Run migrations
 * Click on "Admin" in the sidebar and select "Census Areas"
 * Click around to each view and make sure they're all visible and connected in intuitive ways consistent with the rest of the site
* Navigate to a submitted survey detail page, add a new ACS-compatible chart, and check that a link to the CensusArea list view is displayed in the help text under that field


beamalsky:
@jeancochrane I didn't add tests here yet because I'm not sure how #101 will affect them, but I could add some basic tests here if you think appropriate.

surveys/forms.py
```diff
@@ -6,11 +6,10 @@
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import Submit
 
-from pldp.models import Agency, Location, LocationArea, LocationLine, Study, Survey, \
-                        StudyArea
+import pldp.models as pldp_models
```
jeancochrane: Nice refactor!

PR #183, opened 2019-06-19
Fix missing submit button on user signup page

### LINKED ISSUE #182: User signup page is missing its submit button!
Screen Shot 2019-06-19 at 2 15 37 PM

beamalsky:
## Overview

Closes #182. Adds back a missing submit button on the user signup form.

## Testing Instructions

 * Log out 
* Click "Create new account" in the sidebar
* Create a new account and make sure you can log in with it!


beamalsky:
How do you think UCD should add staff to the app? 

My thinking is that depending on who's able to add new accounts, this signup view could be either public or visible only to admins.

jeancochrane:
I see, I didn't catch that this was intended to be an admin-only view -- that makes more sense.

The simplest solution would be to have admins use the the Django admin interface to manage users (`/admin/users/justspacesuser/`). That way we don't have to build out all of the supporting views that go along with user management (deleting users, changing passwords, adjusting permissions, etc). That would be my preference, unless we've had a conversation in the past indicating that custom user management is an important requirement.

beamalsky:
I see your point here. At one point we'd intended to avoid the admin interface entirely, but I think you're right that this late in the game it makes sense to use those existing views. I'll keep that in mind as I do permissions and refigure the sidebar to direct admins to the Django views to manage users.

jeancochrane:
Cool, sounds good to me!

beamalsky:
Accidentally committed my permissioning system work here first; that will proceed in #185. Closing this!

Issue #66, opened 2019-03-27
About page


reginafcompton:
UCD does not have a strong sense of the intended user base for the public-facing site.

@jeancochrane suggested including a "call out" on the public site, i.e., something that invites visitors to contact such-and-such to get more information about the project, specific survey data, etc. This mechanism could help UCD get a better sense of the user base (which could inform future iterations of the project).

beamalsky:
I'm thinking maybe we can return to this concept, since we're not making a public-facing site. Instead of directing users to the login screen immediately upon visiting the site, they could be sent to the site's "About" page with text from UCD about the project.

PR #185, opened 2019-06-19
Permissioning system

beamalsky:
## Overview

Closes #172. This PR implements three classes of users:

1. `Field users` can run published surveys.
2. `Staff users` have all the permissions of field users. They can also create, edit, and delete Agencies, Studies, Study Areas, Surveys, and Locations. They can publish surveys so they can be run by field users. Staff users can also view all collected data and design data visualizations on the collected data pages.
3. `Superusers` have all the permissions of staff users. They can also create, edit, and delete other users. Only superusers can set and change user passwords. **All superusers should also have staff status.**

@jeancochrane two notes:

1. I put in basic test coverage, but it could certainly be more extensive. Do you think that's worth doing here? 
2. Selecting both "Staff status" and "Superuser status" is confusing, and I'm not sure how best to explain how this works for users without creating new templates. What do you think?

## Testing Instructions

 * If you don't already have one, create a new superuser from the command line
* Once logged in as that superuser, go to `http://127.0.0.1:8000/admin/users/justspacesuser/`
* Create two new users: `fielduser`, who is neither a staff user or a superuser, and `staffuser`, who is both
* Log in as `fielduser`. Make sure that all you can do is select and run surveys
* Navigate to `http://127.0.0.1:8000/agencies` or another protected page. Make sure you're redirected to a login screen
* Log in as `staffuser`. Make sure you can see and do all of the above actions. 
* Create a new user as `staffuser`. Make sure that behaves as expected.


surveys/templates/partials/sidebar.html
```diff
@@ -50,19 +54,16 @@ <h3><b>Just Spaces</b></h3>
...
+                <a href="/">
                     <i class="fas fa-question"></i>
-                    FAQ
-                </a>
-            </li>
-            <li>
-                <a href="#">
-                    <i class="fas fa-paper-plane"></i>
-                    Contact
+                    About
```
jeancochrane: As long as we're getting rid of pages, is there any reason to preserve this `About` pseudo-link?

surveys/urls.py
```diff
@@ -9,6 +9,7 @@
 from django.urls import path
 from django.utils.translation import ugettext_lazy as _
 from django.contrib.auth.decorators import login_required
+from django.contrib.admin.views.decorators import staff_member_required
```
jeancochrane: Right now `staff_member_required` will redirect you to the Django admin login interface instead of our custom login interface. Could we try using the `login_url` param? See: https://stackoverflow.com/a/42038839

jeancochrane:
Two more notes that I forgot:

1. Could we update the README to include some brief background on the two types of users, and how to make them? I don't think we need to go as far as e.g. screenshots, but I think the background you provided in this PR will be broadly useful to anyone looking at this repo.

2. Since we're using the Django Admin interface for user management, It might make sense to disable all of the model editing sections on `/admin/`, other than `User`:

Screen Shot 2019-06-26 at 4 46 20 PM

We should be able to do this by [unregistering them](https://books.agiliq.com/projects/django-admin-cookbook/en/latest/remove_default.html).

surveys/templates/partials/sidebar.html
beamalsky: I was thinking we could bring this page to the attention of UCD and ask them for text, and having this be a very simple public-facing site for users without a login. See https://github.com/datamade/just-spaces/issues/66#issuecomment-503251517

surveys/urls.py
beamalsky: Great point!

beamalsky:
@jeancochrane ready for your eyes again!

beamalsky:
One thing worth noting here is that field users have no way to change their own passwords—staff users will have to change it for them. I don't think we need to implement unless UCD wants it, but might be worth asking directly!

surveys/templates/partials/sidebar.html
jeancochrane: Sounds fine to me! Added it to our agenda for our next call.

README.md
```diff
@@ -98,5 +98,12 @@ If you'd like to refresh ACS data, there are a few steps you'll need to take:
 
 The `import_data` management command will update data if it already exists, so you shouldn't experience data loss during import. Still, we recommend that you practice caution and make a backup.
 
+## Permissioning
+
+This app employs two classes of users:
+
+1. `Staff users` have full access. They can run surveys as well as create, edit, and delete Users, Agencies, Studies, Study Areas, Surveys, and Locations. They can publish surveys so they can be run. Admins can also view all collected data and design data visualizations on the collected data pages. **To create a new staff user, both "Staff status" and "Superuser status" should be selected on the second page of the Django admin user creation interface.**
+2. `Field users` can run published surveys.
```
jeancochrane: Your comment about password management was a good one:

> One thing worth noting here is that field users have no way to change their own passwords—staff users will have to change it for them.

It'd be good to add a quick note here explaining this.

surveys/templates/partials/sidebar.html
```diff
@@ -50,19 +54,16 @@ <h3><b>Just Spaces</b></h3>
                     <li>
                         <a href="{% url 'census-areas-list' %}">- Census Areas</a>
                     </li>
+                    <li>
+                        <a href="/admin/users/justspacesuser/">- Users</a>
+                    </li>
```
jeancochrane: We chatted IRL and agreed we'd restrict this link to `is_superuser` users.

surveys/urls.py
```diff
@@ -22,88 +24,84 @@
...
+        staff_required_custom_login(fobi.views.edit_form_element_entry),
         name='fobi.edit_form_element_entry'),
 
     # Delete survey element
     url(_(r'^elements/delete/(?P<form_element_entry_id>\d+)/$'),
-        view=fobi.views.delete_form_element_entry,
+        staff_required_custom_login(fobi.views.delete_form_element_entry),
         name='fobi.delete_form_element_entry'),
 
     # Create survey handler
```
jeancochrane: Looks like the following views (add, edit, and delete survey handlers) are not protected, do we need to protect those?

beamalsky: Good catch! They're never exposed (Fobi offers the ability to let users add different data handlers. Instead, we've written a custom one and add it automatically to new surveys) so it's not a huge risk, but won't hurt!

users/admin.py
```diff
@@ -19,10 +21,28 @@ class Meta(UserChangeForm.Meta):
...
+        1. <b>Field users</b> can run published surveys.<br /> \
+        2. <b>Staff users</b> have all the permissions of field users. They can \
+            also create, edit, and delete Agencies, Studies, Study Areas, \
+            Surveys, and Locations. They can publish surveys so they can \
+            be run by field users. Staff users can also view all collected \
+            data and design data visualizations on the collected data pages.<br /> \
+        3. <b>Superusers</b> have all the permissions of staff users. They can \
+            also create, edit, and delete other users. Only superusers can set \
+            and change user passwords. <b>All superusers should \
+            also have "Staff status" selected below</b>.<br />'
```
jeancochrane: This copy is super helpful! I think one way we could make it even better would be to explain specifically which buttons need to be checked for each user type, similar to what you've hinted at with superusers. Here's how I would phrase it:

> This app employs three classes of users:
>
> 1. `Field users` can run published surveys. **To create a field user, check 'Active' below.**
> 2. `Staff users` have all the permissions of field users. They can also create, edit, and delete Agencies, Studies, Study Areas, Surveys, and Locations. They can publish surveys so they can be run by field users. Staff users can also view all collected data and design data visualizations on the collected data pages. **To create a staff user, check 'Active' and 'Staff status' below.**
> 3. `Superusers` have all the permissions of staff users. They can also create, edit, and delete other users. Only superusers can set and change user passwords. **To create a superuser, check all the boxes below.**

beamalsky: Looks great! I'll modify and then merge this.

PR #191, opened 2019-07-02
Chart colors

### LINKED ISSUE #188: ACS charts can choose the same color as primary_source
Occasionally when adding or removing a CensusArea from a chart, the DAD will choose the same color for the ACS layer as the `primary_source`:

image001

I'm guessing this is because Highcharts only knows the colors of the series that it's adding at a given time, and not necessarily the series that have already been added.

See if there's a way we can override this behavior to instruct Highcharts not to use colors that have already been displayed on the chart.

beamalsky:
## Overview

This PR fixes the bug described in #188: `ACS charts can choose the same color as primary_source`.

To get around this, I set `primary source` to a consistent color ("Just Spaces mid-green"), and give the ACS series a set of colors to choose from. These can't repeat, as ACS series are torn down and recreated whenever they change. I think this also has the advantage of consistently visually distinguishing collected data from ACS data. If this project is extended to handle multiple series of collected data on the same chart this will have to be updated, but my understanding is that that's not on the horizon for the current scope. 

In the process I've also added a color palette for the charts to bring it more in line with the rest of the site and away from the HighCharts defaults!

It should look like this:

Screen Shot 2019-07-02 at 4 14 16 PM

This also modifies `before_install.sh` to handle blackbox being updated on the server, as documented at https://github.com/datamade/deploy-a-site/pull/69

Closes #188 

## Testing Instructions

 * Create a new chart with an ACS-eligible question
* Add and remove some CensusAreas. Make sure the colors never repeat!

### DIFF
--- a/surveys/static/js/charts.js
+++ b/surveys/static/js/charts.js
@@ -540,7 +540,7 @@ function binValue(value, bins) {
  * @param {Array} a items An array containing the items.
  */
 function shuffle(a) {
-    for (let i = a.length - 1; i > 0; i--) {
+    for (var i = a.length - 1; i > 0; i--) {
         const j = Math.floor(Math.random() * (i + 1));
         [a[i], a[j]] = [a[j], a[i]];
     }
--- a/surveys/static/js/charts.js
+++ b/surveys/static/js/charts.js
@@ -69,7 +69,7 @@ ChartHelper.prototype.loadChart = function(chartId, chartTitle, dataSourceId) {
   var series = [{
     name: this.getSeriesName(),
     data: [],
-    color: '#e94e0c',
+    color: '#a1bfa2',
   }];
   chartData.forEach(function(data) {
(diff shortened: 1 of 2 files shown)


surveys/static/js/charts.js
```diff
@@ -517,3 +534,15 @@ function binValue(value, bins) {
...
+ * Shuffles array in place. ES6 version
+ * @param {Array} a items An array containing the items.
+ */
+function shuffle(a) {
+    for (let i = a.length - 1; i > 0; i--) {
+        const j = Math.floor(Math.random() * (i + 1));
+        [a[i], a[j]] = [a[j], a[i]];
+    }
+    return a;
+}
```
jeancochrane: At this point `let` and `const` are pretty darn close to full browser support with [~94% of the market](https://caniuse.com/#search=let), but they're still not quite universally supported. Since we don't have a transpiler set up for this app, let's go with the safer `var`.

PR #192, opened 2019-07-03
Support charts for ZIP code questions in the DAD

### LINKED ISSUE #184: Wire up Location questions to chart builder
Currently the 4 premade location questions can't be visualized in the chart builder. I think that makes sense for the intersection questions, as they're free text responses, but I think it'll be important to UCD to be able to visualize the distribution of zip codes.

As far as I understand it, `location_zip_home` and `location_zip_work` don't fit into any of the current question types and will need something new. I think what we want is similar to COUNT_TYPES, but a distribution instead of a median. 

@jeancochrane, would you be up for taking this on, since you built this part of the tool? I'm also happy to but it would take me longer to get acquainted with it.

jeancochrane:
## Overview

Closes #184.

Support simple charts for ZIP code location questions in the data analysis designer, showing the distribution of answers:

Screen Shot 2019-07-03 at 12 07 30 PM

## Testing Instructions

* Create an Intercept survey with work/home ZIP code location questions and run it a few times
* Navigate to `View collected data` for your survey and confirm you can create charts out of your ZIP code location questions

### DIFF
--- a/fobi_custom/plugins/form_elements/fields/types.py
+++ b/fobi_custom/plugins/form_elements/fields/types.py
@@ -7,8 +7,8 @@ from fobi.utils import get_registered_form_element_plugins
 COUNT_TYPES = [
     'temperature_c', 'decimal', 'float', 'integer'
 ]
-# "Distribution" types represent a single value, and will be displayed as the
-# as percentiles (a distribution) of all recorded values. This is useful since
+# "Distribution" types represent a single value, and will be displayed as
+# percentiles (a distribution) of all recorded values. This is useful since
 # these types are expected to have a small set of potential responses.
 DISTRIBUTION_TYPES = ['location_zip_home', 'location_zip_work']
 # "Observational" types represent counts for multiple choices, and will
--- a/fobi_custom/plugins/form_elements/fields/types.py
+++ b/fobi_custom/plugins/form_elements/fields/types.py
@@ -7,6 +7,10 @@ from fobi.utils import get_registered_form_element_plugins
 COUNT_TYPES = [
     'temperature_c', 'decimal', 'float', 'integer'
 ]
+# "Distribution" types represent a single value, and will be displayed as the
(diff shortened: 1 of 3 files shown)


fobi_custom/plugins/form_elements/fields/types.py
```diff
@@ -7,6 +7,10 @@
 COUNT_TYPES = [
     'temperature_c', 'decimal', 'float', 'integer'
 ]
+# "Distribution" types represent a single value, and will be displayed as the
+# as percentiles (a distribution) of all recorded values. This is useful since
```
beamalsky: Looks like you've got an extra "as the" here

surveys/static/js/charts.js
```diff
@@ -199,6 +202,34 @@ ChartHelper.prototype._getCountChartData = function(dataSourceId, chartTitle) {
...
+        break;
+      }
+    }
+    if (!categoryFound) {
+      // Insert an entry for this count in the chartData array
+      chartData.push([savedData, 1]);
+    }
+  }
+  var aggFunc = percentiles;
+  return this._getChartData(dataSourceId, initChartDataFunc, castFunc, updateChartDataFunc, aggFunc);
```
beamalsky: What do you think about adding something here to order the savedData? I'd expect the x axis here to go in ascending order:

Screen Shot 2019-07-03 at 2 38 14 PM

beamalsky:
@jeancochrane just bumping this to make sure it's on your radar!

jeancochrane:
Thanks @beamalsky! I haven't forgotten, this is first on my list when I start JS development this week.

surveys/static/js/charts.js
jeancochrane: Great point! Added sorting in https://github.com/datamade/just-spaces/pull/192/commits/d48648461eb70197646edd9ea775c01c4c3bcf58.

PR #194, opened 2019-07-03
Templated surveys

### LINKED ISSUE #90: Templated surveys

beamalsky:
## Overview

Closes #90. This PR creates a pipeline for storing and importing survey templates, then allowing users to choose one when creating a survey. It also creates two example templated surveys. Real templates will be created later by UCD on the staging site and incorporated in #195.

**Note:** As this stands, `python manage.py import_survey_templates` will have to be run manually on the staging server. It cannot be added to `after_install` without also adding a user, which I don't think we want to do automatically right now.

## Testing Instructions

* Run migrations
* Import survey templates: `python manage.py import_survey_templates`
* Start creating a new survey at `localhost:800/surveys/create/`. Make sure the `Survey templates` field contains some options.
* Create the survey! Do you see some templated questions already loaded?


beamalsky:
The Travis build on the last commit passed: https://travis-ci.com/datamade/just-spaces/jobs/214183963

Not sure why it's not connecting here!

surveys/views.py
```diff
@@ -316,8 +319,31 @@ def form_valid(self, form):
                                'collect_data'
                                )
 
+        # add survey questions from template
+        template = form.cleaned_data['survey_template']
+
+        if template:
+            template_questions = fobi_models.FormElementEntry.objects.filter(form_entry=template)
```
jeancochrane: How much of an extra lift would it be to add some validation logic to ensure that only intercept surveys use intercept templates, and vice versa? I'm thinking of potentially two changes:

1. Some conditional logic in the survey creation form that shows/hides options depending on whether the form type was selected as `intercept` or `observational`
2. Backend validation on the `SurveyCreateForm.clean_survey_template()` method to make sure the `type` of the survey matches the `type` of the template (as a safeguard in case the frontend conditional formatting fails, e.g. if the user selects the template before they select the survey type)

beamalsky: Added backend validation! I had thought about this before and added the type of the template in parenthesis to tip users off. Conditional logic (#1) is a heavier lift because it will involve writing JavaScript in the template. If it's ok with you, I don't think that's a high need.

surveys/forms.py
```diff
@@ -165,14 +165,34 @@ def __init__(self, *args, **kwargs):
...
 
+    def clean_survey_template(self):
+        template = self.cleaned_data['survey_template']
+        survey_type = self.cleaned_data['type']
+
+        if template and (template.surveyformentry.type != survey_type):
+            if survey_type == 'intercept':
+                raise forms.ValidationError("Make sure you choose an intercept template to create a new intercept survey!")
+            if survey_type == 'observational':
+                raise forms.ValidationError("Make sure you choose an observational template to create a new observational survey!")
```
jeancochrane: We could simplify this a little bit to generalize to both cases:

```python
if template and (template.surveyformentry.type != survey_type):
    raise forms.ValidationError(
        "Make sure you choose an {} template to create a new {} survey!".format(survey_type)
    )
```

beamalsky: Oo true!

beamalsky: Needed a slight syntax change:

```python
if template and (template.surveyformentry.type != survey_type):
	raise forms.ValidationError(
	"Make sure you choose an {0} template to create a new {0} survey!".format(survey_type)
	)
```

PR #196, opened 2019-07-08
Filter for deactivated models

### LINKED ISSUE #176: Deactivated records should not show up in Create dropdowns
Make sure that deleted records don't show up in the following dropdown fields: 

- Study and Location in "Create new Survey"
- Agency in "Create new Study"

beamalsky:
## Overview

Closes #176. A quick PR that fixes something that might appear like a bug! Makes sure deactivated records of the following types won't show up in dropdowns:

- Study and Location in "Create new Survey"
- Agency in "Create new Study"

## Testing Instructions

 * Go to "Create new Survey". Deactivate one of the Study and Location records each and make sure they disappear.
* Do the same for Agency on the "Create new Study" view

### DIFF
--- a/surveys/forms.py
+++ b/surveys/forms.py
@@ -146,7 +146,10 @@ class StudyCreateForm(JustSpacesForm):
         super(StudyCreateForm, self).__init__(*args, **kwargs)
         self.create_default_helper()
         self.fields['areas'].widget.attrs['class'] = 'basic-multiple'
+
         self.fields['agency'].initial = pldp_models.Agency.objects.first()
+        self.fields['agency'].queryset = pldp_models.Agency.objects.filter(is_active='t')
+
         self.fields['title'].required = True
 
     class Meta:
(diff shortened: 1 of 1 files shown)


surveys/forms.py
```diff
@@ -146,7 +146,10 @@ def __init__(self, *args, **kwargs):
         super(StudyCreateForm, self).__init__(*args, **kwargs)
         self.create_default_helper()
         self.fields['areas'].widget.attrs['class'] = 'basic-multiple'
+
         self.fields['agency'].initial = pldp_models.Agency.objects.first()
+        self.fields['agency'].queryset = pldp_models.Agency.objects.filter(is_active='t')
```
jeancochrane: Forgot to raise this in the course of reviewing #194: What's the advantage of filtering on `is_active='t'` instead of `is_active=True`?

beamalsky: It's just the style I've been using! That's how it shows up in psql so it made sense to me, and then I kept it consistent. Does it make a difference?

jeancochrane: Gotcha! I couldn't find any docs on what specific values would evaluate to `True` when filtering on a BooleanField in the Django ORM but it makes sense that `'t'` would work. From a stylistic perspective using the `True` boolean makes a bit more sense to me than `'t'` since the ORM is supposed to be an abstraction layer that serializes Postgres records into Python objects, and `True` is the Pythonic boolean. But since `'t'` works in this case I think it's a question of taste.

If you're curious, I found a thread on the Django dev google group discussing this behavior: https://groups.google.com/forum/#!topic/django-developers/f82vcCMrpSs Looks like anything truthy will evaluate to `True`, and anything falsey will evaluate to `False`. Some interesting discussions of edge cases in there.

beamalsky: Makes sense! I'll use `True` next time.

Issue #203, opened 2019-07-11
Add preset Observational race question


jeancochrane:
Implement a preset Observational question for race, using the same categories as the preset Intercept question.

In addition, wire the question up to the DAD.

jeancochrane:
Ask UCD for some copy for this help text.

beamalsky:
Once this is implemented we'll need to add this to the templates too!

PR #205, opened 2019-07-15
Hard coded template surveys

### LINKED ISSUE #195: Hard code UCD survey templates
The last bit of #90. Waiting on UCD to provide these surveys!

beamalsky:
## Overview

Closes #195. This PR hard codes templated surveys as described by Tyler here: https://docs.google.com/document/d/1WGEWdb3OGAGZfvunCiUNmrO7NanYhDBPe14lT1931DE/edit

I put together the JSON template file manually, but this could/should be automated in the future if this project is extended! I'll leave couple notes in-line on changes as well. 

We'll need to add the `race_observational` question to these templates once it's created, which I've noted in #203.

## Testing Instructions

 * Import templates: `python manage.py import_survey_templates`
* Create a new survey with each survey template. Make sure that if they types don't match the form is invalid.
* Make sure you can edit templated survey questions

### DIFF
--- a/fobi_custom/plugins/form_elements/fields/intercept/age_intercept/forms.py
+++ b/fobi_custom/plugins/form_elements/fields/intercept/age_intercept/forms.py
@@ -12,7 +12,6 @@ class AgeInterceptForm(forms.Form, BaseFormFieldPluginForm):
     plugin_data_fields = [
         ("label", "How old are you?"),
         ("name", "name"),
-        ("age", ""),
         ("help_text", ""),
         ("required", ""),
     ]
--- a/surveys/models.py
+++ b/surveys/models.py
@@ -50,7 +50,7 @@ class SurveyFormEntry(FormEntry):
     active = models.BooleanField(default=True)
 
     def __str__(self):
-        return self.name + " (" + str(self.type) + ")"
+        return self.name
 
 
 class CensusBlockGroup(models.Model):
--- a/surveys/survey_template_data/survey_templates.json
+++ b/surveys/survey_template_data/survey_templates.json
@@ -1,23 +1,165 @@
 [
   {
-    "name": "Sample Survey Template Intercept",
-    "type": "intercept",
+    "name": "Event Survey – Observational",
+    "type": "observational",
     "questions": [
       {
-        "plugin_data": "{\"label\": \"What year did you move into your current address?\", \"age\": null, \"help_text\": \"\", \"required\": false}",
-        "plugin_uid": "household_tenure",
+        "plugin_data": "{\"label\": \"How many people do you see in each age range?\", \"detail_level\": \"basic\", \"help_text\": \"\", \"required\": false}",
+        "plugin_uid": "age_observational",
         "position": 1
+      },
+      {
+        "plugin_data": "{\"label\": \"How many people do you see of each gender?\", \"help_text\": \"\", \"required\": false}",
+        "plugin_uid": "gender_observational",
+        "position": 2
+      },
+      {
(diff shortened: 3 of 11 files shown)


fobi_custom/plugins/form_elements/fields/intercept/age_intercept/forms.py
```diff
@@ -12,7 +12,6 @@ class AgeInterceptForm(forms.Form, BaseFormFieldPluginForm):
     plugin_data_fields = [
         ("label", "How old are you?"),
         ("name", "name"),
-        ("age", ""),
```
beamalsky: This wasn't hurting anything but is a stray copy-paste mistake that I've deleted from a couple plugins.

surveys/models.py
```diff
@@ -50,7 +50,7 @@ class SurveyFormEntry(FormEntry):
     active = models.BooleanField(default=True)
 
     def __str__(self):
-        return self.name + " (" + str(self.type) + ")"
```
beamalsky: Since Tyler's template include the `type` in the `name`, and since Fobi restricts templates with the same name created by the same user, I've dropped the `type` here

surveys/views.py
```diff
@@ -405,7 +405,7 @@ class SurveyListEdit(ListView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context['surveys'] = context['surveys'].exclude(active=False).filter(published=False).order_by('-updated')
```
beamalsky: Making sure templates don't show up as editable or runnable surveys!

surveys/models.py
jeancochrane: This seems reasonable to me! We'll just have to remember to include the survey type in the name of any new templates.

surveys/survey_template_data/survey_templates.json
```diff
@@ -1,23 +1,165 @@
...
+        "position": 4
+      },
+      {
+        "plugin_data": "{\"label\": \"What is the zip code of your workplace?\", \"help_text\": \"\", \"required\": false}",
+        "plugin_uid": "location_zip_work",
+        "position": 5
+      },
+      {
+        "plugin_data": "{\"label\": \"What year did you move into your current address?\", \"help_text\": \"\", \"required\": false}",
+        "plugin_uid": "household_tenure",
```
jeancochrane: Tyler specified slightly different wording for these three location questions. Did you have a reason for altering them? I'm open to different wording, I just want to make sure I can communicate to him the reason we changed them.

surveys/views.py
jeancochrane: Nice catch! This line is getting a little long (117 characters), could we add in a linebreak? Maybe something like:

```python
context['surveys'] = context['surveys'].filter(
    active=True, is_cloneable=False, published=False
).order_by('-updated')
```

surveys/views.py
```diff
@@ -418,7 +418,7 @@ class SurveyListRun(ListView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context['surveys'] = context['surveys'].exclude(active=False).filter(published=True).order_by('-updated')
+        context['surveys'] = context['surveys'].filter(active=True, is_cloneable=False, published=True).order_by('-updated')
```
jeancochrane: Ditto here for line length.

surveys/models.py
beamalsky: Yep!

surveys/survey_template_data/survey_templates.json
beamalsky: This is the wording in the built in questions, and he already said we should use that when his differ slightly!

jeancochrane: Oops, got it! Sorry for not reading closer 👍

PR #206, opened 2019-07-16
Add new Observational question for perceived race

### LINKED ISSUE #203: Add preset Observational race question

jeancochrane:
## Overview

Closes #203.

Create a new Observational question for Perceived Race. Wire it up to the DAD and allow the display of ACS race data alongside collected data.

## Demo

Screen Shot 2019-07-16 at 11 34 30 AM

## Testing Instructions

* Run `python manage.py import_survey_templates`
* Spin up the dev server and create or edit an Observational survey
* Add a question for `Observational > Perceived Race`
* Publish the survey and run it a few times
* View the Perceived Race question in the DAD and confirm you can show it alongside ACS data
* Create a new Observational survey from one of the templates and confirm that the Perceived Race question shows up appropriately

### DIFF
--- a/fobi_custom/plugins/form_elements/fields/observational/race_observational/fobi_form_elements.py
+++ b/fobi_custom/plugins/form_elements/fields/observational/race_observational/fobi_form_elements.py
@@ -1,0 +1,31 @@
+from fobi.base import FormFieldPlugin, form_element_plugin_registry
+from fobi_custom.plugins.form_elements.fields.intercept.forms import RACE_CHOICES
+
+from ..widgets import ObservationalWidget
+from ..fields import ObservationalField
+
+from .forms import RaceObservationalForm
+
+
+class RaceObservationalPlugin(FormFieldPlugin):
+    """RaceObservationalPlugin."""
+
+    uid = "race_observational"
+    name = "Perceived Race"
+    form = RaceObservationalForm
+    group = "Observational"  # Group to which the plugin belongs to
+
+    def get_form_field_instances(self, request=None, form_entry=None,
+                                 form_element_entries=None, **kwargs):
+
+        field_kwargs = {
+            'required': self.data.required,
+            'label': self.data.label,
+            'choices': RACE_CHOICES,
+            'widget': ObservationalWidget(choices=RACE_CHOICES)
+        }
+
+        return [(self.data.name, ObservationalField, field_kwargs)]
+
+
+form_element_plugin_registry.register(RaceObservationalPlugin)
--- a/fobi_custom/plugins/form_elements/fields/observational/race_observational/forms.py
+++ b/fobi_custom/plugins/form_elements/fields/observational/race_observational/forms.py
@@ -10,9 +10,9 @@ class RaceObservationalForm(forms.Form, BaseFormFieldPluginForm):
     """RaceObservationalForm."""
 
     plugin_data_fields = [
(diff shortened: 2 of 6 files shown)


fobi_custom/plugins/form_elements/fields/observational/race_observational/fobi_form_elements.py
```diff
@@ -0,0 +1,31 @@
+from fobi.base import FormFieldPlugin, form_element_plugin_registry
+from fobi_custom.plugins.form_elements.fields.intercept.forms import RACE_CHOICES
```
jeancochrane: Importing `RACE_CHOICES` saves time and aligns things with the corresponding Intercept question, but it has the downside of adding `Other` as a category to the question. Methodologically this seems kind of weird to me (we're already making surveyors guess the person's race, so what does `Other` mean in that context?) but I might be overthinking it. Another important consideration is that keeping  the `Other` category also makes it easier to line things up with the ACS data.

fobi_custom/plugins/form_elements/fields/observational/race_observational/forms.py
```diff
@@ -0,0 +1,42 @@
...
+from fobi.base import BaseFormFieldPluginForm, get_theme
+
+theme = get_theme(request=None, as_instance=True)
+
+
+class RaceObservationalForm(forms.Form, BaseFormFieldPluginForm):
+    """RaceObservationalForm."""
+
+    plugin_data_fields = [
+        ("label", "How many people of each perceived category do you see? (select all that apply)"),
```
jeancochrane: It'd be nice to move `Select all that apply` to be the default `help_text` for this question, but I couldn't figure out how to prepopulate it after messing with it for a few minutes (I tried the `initial` kwarg in the `CharField` below, but that didn't seem to work).

fobi_custom/plugins/form_elements/fields/observational/race_observational/fobi_form_elements.py
beamalsky: It makes sense to me to keep consistency with the ACS, even if it's a little weird! I think `Other` functions well as a "not sure" option.

fobi_custom/plugins/form_elements/fields/observational/race_observational/forms.py
beamalsky: I'll leave a comment where this should go!

fobi_custom/plugins/form_elements/fields/observational/race_observational/forms.py
```diff
@@ -0,0 +1,42 @@
...
+        ("name", "name"),
+        ("help_text", ""),
```
beamalsky: To prepopulate help text, this should be: 
```("help_text", "Select all that apply."),```

fobi_custom/plugins/form_elements/fields/observational/race_observational/forms.py
beamalsky: To maintain consistency, I'd cut "(select all that apply)" from here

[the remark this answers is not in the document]
jeancochrane: Good catch. I copied and pasted Tyler's wording for the question but I agree that consistency is more important.

PR #207, opened 2019-07-16
Production deployment

### LINKED ISSUE #200: Deploy production site
Follow the guides in https://github.com/datamade/deploy-a-site/ to spin up a production server for the application.

We're waiting on UCD for the domain name and DNS, but we should be able to move forward with everything else.

beamalsky:
## Overview

Connects #200. Just waiting on a production domain once UCD picks one! 

I updated all files in blackbox so they all show up as changed, though mostly those weren't actual changes.

## Testing Instructions

 * Check out [justspaces-production.datamade.us](https://justspaces-production.datamade.us/) (it has the same `testuser` account as staging)!


scripts/after_install.sh
```diff
@@ -45,14 +45,15 @@ psql -U postgres -d just-spaces -c "CREATE EXTENSION IF NOT EXISTS postgis"
 
 # OPTIONAL Run migrations and other management commands that should be run with
 # every deployment
-$VENV_DIR/bin/python $PROJECT_DIR/manage.py migrate
-$VENV_DIR/bin/python $PROJECT_DIR/manage.py createcachetable
-$VENV_DIR/bin/python $PROJECT_DIR/manage.py collectstatic --no-input
+sudo -H -u datamade $VENV_DIR/bin/python $PROJECT_DIR/manage.py migrate
```
beamalsky: This was needed to run migrations but isn't reflected in current SOP here: https://github.com/datamade/deploy-a-site/blob/master/templates/django/%7B%7Bcookiecutter.directory_name%7D%7D/scripts/after_install.sh#L48

I'm honestly not sure why this ever works. Do you think this is worth opening an issue in `deploy-a-site`?

data/.gitignore
```diff
@@ -5,9 +5,3 @@ intermediate/acs/*.csv
 
 # Extra shapefile data
 intermediate/shapefiles/*.shp
-*/shapefiles/*.cpg
```
jeancochrane: We can still go ahead and ignore `intermediate/` shapefiles, I think.

scripts/after_install.sh
jeancochrane: Yeah, this is pretty weird behavior. Do you mind opening an issue in `devops` to track this?

scripts/after_install.sh
```diff
@@ -45,14 +45,15 @@ psql -U postgres -d just-spaces -c "CREATE EXTENSION IF NOT EXISTS postgis"
...
+sudo -H -u datamade $VENV_DIR/bin/python $PROJECT_DIR/manage.py createcachetable
+sudo -H -u datamade $VENV_DIR/bin/python $PROJECT_DIR/manage.py collectstatic --no-input
 
 # Echo a simple nginx configuration into the correct place, and tell
 # certbot to request a cert if one does not already exist.
 # Wondering about the DOMAIN variable? It becomes available by source-ing
 # the config file (see above).
+sudo apt install python-certbot-nginx
```
jeancochrane: Does this need to be added to our deployment templates?

scripts/after_install.sh
beamalsky: Opened! https://github.com/datamade/devops/issues/105

scripts/after_install.sh
beamalsky: I think this should be added to the server configs in this line: https://github.com/datamade/deploy-a-site/blob/cbacb18a0b8f612a76b7e4877576cb4e124e8a86/templates/server/%7B%7Bcookiecutter.directory_name%7D%7D/scripts/after_install.sh#L16

Does that seem right to you?

beamalsky:
`data/final/acs/census_observations.csv` isn't deleted, just shorter! It seems to work for me locally—can you think of a reason why the `import_data` command would produce a shorter file since the last time it was under version control?

scripts/after_install.sh
jeancochrane: That's the right place, but it looks like it's on there already:

```
packages='apt-transport-https ca-certificates curl software-properties-common nginx supervisor python3 python3-pip python3-venv certbot python-certbot-nginx'
```

Any idea why it wasn't working for you?

jeancochrane:
> data/final/acs/census_observations.csv isn't deleted, just shorter! It seems to work for me locally—can you think of a reason why the import_data command would produce a shorter file since the last time it was under version control?

I checked this diff locally and I think GitHub may be parsing it incorrectly. My local Git client shows each row with a different carriage-return added to the end of it (`^M`), which I don't think should make a difference. Not sure why this happened but if it's importing properly on production then I'm fine accepting it.

scripts/after_install.sh
beamalsky: Because I copy and pasted from here instead!! https://github.com/datamade/deploy-a-site/blob/master/Basic-server-setup.md

I can put together a PR on that repo to reflect that change

data/.gitignore
jeancochrane: We should also remove the files from version control before merging, in addition to ignoring them.

scripts/after_install.sh
jeancochrane: Ah, gotcha!

PR #208, opened 2019-07-22
Add about page

### LINKED ISSUE #199: Create About page
Once we have content from UCD, turn it into an About page that is visible to non-logged-in users.

beamalsky:
## Overview

Closes #199. Creates an About page using text from this UCD document: https://docs.google.com/document/u/1/d/18MsGjBQjqoLFtKZYexaJC_v1N18yAKKVuJJameOzGHI/edit?usp=sharing

About page should be accessible without a login and include pictures taken from UCD's site. This PR also incorporates the body text font from UCD's central branding. 

I grabbed some pictures from UCD's website for this. It would be nice, but more time-intensive, to populate this page with screengrabs from the site itself to show what it looks like. 

Screen Shot 2019-07-22 at 2 08 06 PM

## Testing Instructions

 * Run the site locally and navigate to `http://localhost:8000/`
* Look over the about page and make sure everything's in order


tests/test_views.py
```diff
@@ -9,6 +9,14 @@
 from fobi_custom.plugins.form_elements.fields import types as fobi_types
 
 
+@pytest.mark.django_db
+def test_study_area_create():
```
jeancochrane: We'll want to switch this to something like:

```suggestion
def test_home_page_loads():
```

beamalsky: Thx 😅

PR #213, opened 2019-07-29
Add create account button

### LINKED ISSUE #211: Add "Create account" button to the sidebar
From Tyler's email:

> When logged out, the left column shows an “About” button and a “Login” button. Could you add a “Create Account” button that links to the google form?

beamalsky:
## Overview

Closes #211. Adds link to UCD's "Create account" form on the sidebar when user is not logged in. @jeancochrane do you think they want this visible after login as well?

## Testing Instructions

 * Run the site locally and look at the sidebar. Click the "Create account" button. Does it work?

### DIFF
--- a/surveys/static/css/custom.css
+++ b/surveys/static/css/custom.css
@@ -141,7 +141,7 @@ h5 {
 }
 
 .sidebar ul li a:hover {
-  padding: 5px 25px;
+  padding: 5px 27px;
 }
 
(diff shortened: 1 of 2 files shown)


beamalsky:
Good point, I was thinking about this too! How about the below, reflected in my last commit:

Screen Shot 2019-07-29 at 5 36 19 PM

PR #218, opened 2019-08-15
Restrict list views based on the user's Agency

jeancochrane:
## Overview

Restrict list views based on the user's Agency. If the user has no assigned Agency, show them everything; likewise, if the models in question have no assigned Agency, show them to all users.

## Testing Instructions

* Confirm tests pass on Travis
* Confirm that your default user has no Agency assigned
* Confirm that your default user can continue to see all objects
* Create a new user assigned to a specific Agency and log in as that user
* Navigate to all list views and confirm that your user can only see objects created by your Agency
* Create a new Survey with your user
* Check the database and confirm that the Survey is assigned to your agency, and not to the default agency
* Navigate to http://localhost:8000/census-areas/create
* Confirm you can save an Agency for the CensusArea


jeancochrane:
Good points @beamalsky! I believe I addressed them all in https://github.com/datamade/just-spaces/pull/218/commits/ec37b193585a0fc37fad3a85491bc43c6685f595.

Agreed that restricting the available Agencies in Agency dropdowns in e.g. `CensusAreaCreate` and `LocationCreate` isn't necessarily in the scope of this PR, but restricting the Studies available in `SurveyCreate` seems important. Great catch there -- I added this functionality along with a test.

PR #219, opened 2019-09-25
Make sure SurveyEdit view restricts QuerySet where published=False

jeancochrane:
## Overview

A typo in the `queryset` of `surveys.views.SurveyListEdit` meant that we were filtering the QuerySet to surveys where `published=False`. This is exactly the opposite of what we want! In addition, the corresponding check wasn't thoroughly checking which surveys were returned on the page, so the test was passing incorrectly.

Adjust `SurveyListEdit.queryset` and the corresponding test to fix this bug.

## Testing Instructions

* Make sure tests pass
* Navigate to http://localhost:8000/surveys/create/ and create a new test survey
* On the question edit view, hit `Save and Exit`
* Navigate to http://localhost:8000/surveys/edit/ and confirm your new survey shows up there
* Navigate to http://localhost:8000/surveys/run/ and confirm your new survey does not show up there

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -147,11 +147,6 @@ def test_location_detail(client, user_staff, location):
 def test_survey_list_edit(client, user_staff, survey_form_entry,
                           survey_form_entry_inactive, survey_form_entry_observational,
                           survey_form_entry_agency_2):
-    # Make sure that the SurveyFormEntry from Agency 2 is not published, to verify
-    # that it doesn't get loaded on the Edit list view.
-    survey_form_entry_agency_2.published = False
-    survey_form_entry_agency_2.save()
-
     client.force_login(user_staff)
     url = reverse('surveys-list-edit')
     response = client.get(url)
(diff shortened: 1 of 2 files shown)


tests/test_views.py
```diff
@@ -147,11 +147,6 @@ def test_location_detail(client, user_staff, location):
 def test_survey_list_edit(client, user_staff, survey_form_entry,
                           survey_form_entry_inactive, survey_form_entry_observational,
                           survey_form_entry_agency_2):
-    # Make sure that the SurveyFormEntry from Agency 2 is not published, to verify
-    # that it doesn't get loaded on the Edit list view.
-    survey_form_entry_agency_2.published = False
-    survey_form_entry_agency_2.save()
```
jeancochrane: This fixture is initialized with `published=False` so the update here is redundant.

tests/test_views.py
```diff
@@ -160,6 +155,7 @@ def test_survey_list_edit(client, user_staff, survey_form_entry,
 
     assert response.status_code == 200
     assert len(surveys) == 1
+    assert surveys[0] == survey_form_entry_observational
```
jeancochrane: Previously the only contents of `surveys` was `survey_form_entry`, which has `is_active=True` and belongs to the user's Agency but has `published=True`.

Issue #221, opened 2019-10-15
Allow user to create CensusAreas for new cities


jeancochrane:
Update the CensusArea create view to allow the user to add CensusAreas for new cities:

1. Create a query param that filters the geographies and centers the map depending on the selected city
2. Create an autocomplete search box for dropdowns that reloads the page with new geographies when the user selects a new city
3. (Optional, depending on usability) Add a button like "Create a new geography in this city" to allow the user to circumvent the city selection process when creating lots of geographies

This will augment the work in #220 and allow new CensusAreas to be useable.

PR #222, opened 2019-10-28
Define interface for selecting Census regions

### LINKED ISSUE #223: AgencyRestrictQuerysetMixin causes object create to fail
@beamalsky noticed in #222 that after creating a new CensusArea, the new CensusArea is not immediately visible in the ListView that the app redirects to you.

After digging into it a little bit more, I was able to determine that this is a problem with the `AgencyRestrictQuerysetMixin` introduced in #218, meaning that it affects all inheritors of that mixin -- including `LocationList`, `StudyList`, `SurveyListEdit`, `SurveyListRun`, and `SurveySubmittedList`.

In each of these cases, we have a `queryset` defined as an attribute of the class, as well as a `get_queryset()` method that uses `AgencyRestrictQuerysetMixin.get_queryset_for_agency()` to restrict the queryset based on the agency, e.g.:

https://github.com/datamade/just-spaces/blob/2c1b312e6e14692eaccadc272a278a04eb6a4b11/surveys/views.py#L528-L535

`get_queryset_for_agency()` in turn filters `self.queryset` to return the final queryset:

https://github.com/datamade/just-spaces/blob/2c1b312e6e14692eaccadc272a278a04eb6a4b11/surveys/views.py#L86-L99

The crux of the problem is that **`self.queryset` is cached in some kind of way that causes it to fail to re-evaluate when a new object is created.** If you drop a breakpoint in `get_queryset_for_agency()` and create a new object, you'll notice that `self.queryset` does not include the object whereas `super(CensusAreaList, self).get_queryset()` does (since it's evaluated dynamically).

In a weird twist, the test client does not exhibit this caching behavior, and testing this behavior explicitly just causes tests to pass as normal.

I don't fully grasp the underlying caching behavior here since [the Django docs](https://docs.djangoproject.com/en/2.2/topics/class-based-views/generic-display/#viewing-subsets-of-objects) don't provide any details. However, I think the right move is to shift these methods away from relying on `self.queryset` and instead call `get_queryset()` dynamically every time we need to filter.

jeancochrane:
## Overview

This PR sketches out the UI for selecting Census regions when creating new CensusAreas. Also, fix a bug where `View.get_queryset_for_agency()` improperly caches the queryset.

Connects #221, closes #223.

## Notes

- I wanted to get this PR in before actually adding new CensusRegions so that we could figure out what the interface should look like. Because of this, the new view won't really show any new regions -- it should simply illustrate what the user flow will look like, with Philadelphia as the only currently-supported region.

- #221 suggests that we might try to create a "shortcut" workflow that will skip the region selection step for power users who only create CensusAreas in one region. I didn't add that as part of this PR yet but I'd be curious to hear if you have thoughts about how we might achieve that -- there doesn't seem to me to be anywhere obvious where it would make sense.

## Testing Instructions

* Run migrations and start dev server
* Log in and navigate to http://localhost:8000/census-areas/
* Hit the `Create new census area` button
* On the new form view, add a name and agency for your area, and choose `Philadelphia` from the Region selector
* Confirm you get redirected to an updated form including the map with census tracts in Philadelphia
* Go back to http://localhost:8000/census-areas/ and edit an existing CensusArea
* Confirm that the Region input is disabled


jeancochrane:
I did a little bit of debugging and was able to figure out the root problem with creating new CensusAreas, which I've captured in https://github.com/datamade/just-spaces/issues/223. Tomorrow I'll add in a fix for it, add `Agency.default_region`, and also add a little spinner to the CensusRegion select form so the user knows to hang tight.

beamalsky:
I pulled down and I'm not seeing an Agency checkbox! 

Screen Shot 2019-10-30 at 2 58 41 PM

jeancochrane:
@beamalsky It should be below the map! Do you see it there?

Issue #220, opened 2019-10-15
Import block groups for new places


jeancochrane:
Tyler is going to provide us with a list of GeoIDs for counties/statistical areas that they want to import for use as CensusAreas. Update the import script to:

1. Bring in new block group shapefiles
2. Load in ACS data for these places

PR #225, opened 2019-11-01
Update data import pipeline for new geographies

jeancochrane:
## Overview

This PR updates the data import pipeline to add new CensusRegions and CensusBlockgroups to the app. It also makes some small tweaks to the display of the CensusArea widget map to allow display of different regions and fix a few bugs.

Once this PR is landed and Tyler has approved the general approach, we can go ahead and add the remaining ~15 cities.

Connects #220. 

## Testing Instructions

* Clean out your local data files with `docker-compose run --rm make clean`
* Grab new data with `docker-compose run --rm make all` (this may take 5-10 minutes)
* Run a fresh import with `python manage.py import_data`
* Run the app and navigate to http://localhost:8000/census-areas/
* Try creating new CensusAreas in the new regions
* Log in as a staff user (non-superuser) and confirm that the Create view for CensusAreas hides the Agency checkbox from you
* Test that the new CensusAreas can be used to filter census data in the DAD


data/.gitignore
```diff
@@ -1,7 +1,9 @@
...
 raw/acs/*.csv
-raw/shapefiles/*.zip
 intermediate/acs/*.csv
 
-# Extra shapefile data
+# Shapefiles
+raw/shapefiles/*.zip
 intermediate/shapefiles/*
+final/shapefiles/*
+final/acs/*
```
jeancochrane: The final ACS data and shapefiles are now so big that it's not feasible to store them under source control anymore. Instead, we'll have to require that new users run the Makefile to set up a dev environment.

surveys/templates/partials/multiselect_geometry_widget.html
```diff
@@ -133,13 +134,13 @@
       if (layer.feature.properties.selected) {
         // Layer was already selected; turn it off and remove it from the list of
         // selected layers
+        resetHighlight(e);
```
jeancochrane: In particularly huge regions (e.g. Atlanta) the methods `SelectedLayers.pop()` and `SelectedLayers.push()` can have a slight lag, since the JavaScript needs to iterate the selectbox options and select/deselect options as appropriate. Because of this, I moved the functions that indicate UI changes (`resetHightlight()` and `layer.setStyle()` to the top of the block to make the user latency as low as possible.

surveys/templates/partials/multiselect_geometry_widget.html
```diff
@@ -118,6 +118,7 @@
           e.preventDefault();
           jsonLayer.eachLayer(function(layer) {
             layer.setStyle(layerStyle);
+            layer.feature.properties.selected = false;
```
jeancochrane: The lack of this line was causing styles to go a little haywire after the user hit "Clear all". Because we weren't setting `selected = false` on the actual `layer` object, the layer would be deselected from the form, but it would refuse to be styled correctly by `resetHightlight()` on user mouseout.

jeancochrane:
Quick note: Tyler expressed a desire to hide the agency restriction checkbox from anyone who wasn't a superuser, so I snuck that into this PR. Testing instructions updated!

jeancochrane:
Great comments, thanks @beamalsky! I added the requested documentation so I'll pull the PR in now. Agreed that I don't like the nonstandard architecture of the data import process, we can revisit that next time it seems like we need to make changes to the import.

PR #228, opened 2019-11-06
Add new regions to data import

jeancochrane:
## Overview

Add a few new regions to finalize the new geographic import pipeline drafted in #225.

In addition, add some documentation explaining how to do additional imports like this one in the future.
 
### Redo the data import

There are only a few commands you need to run to redo the data import, but they can take a long time to complete. Get set up on something else while you wait.

* Change to the data directory: `cd data`
* Clean out old files: `docker-compose run --rm make clean`
* Create new source data: `docker-compose run --rm make` (this will take 20-30 minutes)
* Change back to the project directory: `cd ..`
* Import the new data `./manage.py import_data` (this will take 20-30 minutes)

### Test the new regions

* Run the development server: `./manage.py runserver`
* Log in and navigate to http://localhost:8000/census-areas/region/
* Use the region selector to load block groups for each region and confirm they look OK
    * Note: Two regions (New York and LA) contain so many block groups that the map widget feels slow on my machine. I'm going to raise this with Tyler and see how he wants to proceed -- I think an easy first step would be to restrict the block groups to NYC and LA instead of including the surrounding counties. If you notice any other regions that you think are unacceptably slow, let me know and I can raise those too.

### DIFF
--- a/data/README.md
+++ b/data/README.md
@@ -55,3 +55,13 @@ docker-compose run --rm make clean
 Next, adjust the global variable `YEAR` in `Makefile` to correspond to the year you'd like to retrieve ACS estimates for.
 
 Finally, rerun the pipeline with `docker-compose run --rm make`. This will create new files in the `data` directory that you can use to import into the app.
+
+## Adding a new Region
+
+If you'd like to add a new CensusRegion, here are the steps you should follow:
+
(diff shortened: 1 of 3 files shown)


jeancochrane:
Pulling this in so that we can show Tyler something today. If we encounter issues on staging we can open a new PR.

PR #229, opened 2019-11-08
Restrict DAD CensusArea choices by user's Agency

### LINKED ISSUE #227: SurveyChartForms do not restrict CensusAreas by Agency
If you view a chart for collected data, e.g. on https://justspaces.datamade.us/surveys/submitted/4, you'll notice that the `Compare to Census` widget does not filter `CensusAreas` by Agency:

Screen Shot 2019-11-06 at 12 27 05 PM

We should update the form initialization code to implement this restriction:

https://github.com/datamade/just-spaces/blob/23c73381d4aac66f12d10a8b6ded2013d1150947/surveys/forms.py#L301-L317

jeancochrane:
## Overview

While finishing up the data import in #228, I noticed that SurveyChartForms were not properly restricting CensusAreas based on the user's Agency or `is_active` status. This PR fixes that small bug.

Closes #227. 

## Testing Instructions

* Make sure you have two Agencies with non-superuser staff members in each Agency
* Log in as one of your staff members and create a new CensusArea at http://localhost:8000/census-areas/region/
* Navigate to a submitted survey detail page at http://localhost:8000/surveys/submitted, load a chart, and confirm that the CensusArea you created is in the dropdown
* Log in as the staff member in the other Agency, navigate to a submitted survey detail page at http://localhost:8000/surveys/submitted, load a chart, and confirm that the CensusArea you created is **not** in the dropdown
* Log in as a superuser, navigate to a submitted survey detail page at http://localhost:8000/surveys/submitted, load a chart, and confirm that the CensusArea you created **is** in the dropdown
* As the superuser, navigate to http://localhost:8000/census-areas/ and delete the CensusArea you just created
* Navigate back to the submitted survey detail page and confirm that the CensusArea is no longer available in the chart selector dropdown

### DIFF
--- a/surveys/forms.py
+++ b/surveys/forms.py
@@ -318,7 +318,7 @@ class SurveyChartForm(forms.ModelForm):
         self.fields['primary_source'].widget.choices = source_choices
 
         # Restrict CensusAreas by the user's Agency
-        census_areas = survey_models.CensusArea.objects.all()
+        census_areas = survey_models.CensusArea.objects.filter(is_active=True)
         if user.agency is not None and not user.is_superuser:
             census_areas = census_areas.filter(
                 Q(agency=user.agency) | Q(agency__isnull=True)
--- a/surveys/forms.py
+++ b/surveys/forms.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 
 from django import forms
-
+from django.db.models import Q
 from leaflet.forms.widgets import LeafletWidget
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import Submit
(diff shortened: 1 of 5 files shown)


jeancochrane:
@beamalsky I dug into it a little bit more and figured out that we weren't properly exempting superusers from the Agency restriction check in the CensusArea list view in the case where the suepruser was assigned an Agency. This helps explain why I was seeing some CensusAreas in the chart creator but not in the list view. I pushed up https://github.com/datamade/just-spaces/pull/229/commits/df17af7fedc0ad4404f01b6de909f6091d17e69a to adjust this and make sure the tests covered this case.

However, this still doesn't help explain the behavior you saw. In particular, I'm wondering why your superuser, who had a different Agency from the staff user that created the CensusArea in question, was able to see the CensusArea in the list view that they then deleted. Given the bug I fixed in  https://github.com/datamade/just-spaces/pull/229/commits/df17af7fedc0ad4404f01b6de909f6091d17e69a, shouldn't this CensusArea have been hidden from that superuser?

beamalsky:
Following your testing instructions above again, this works as you expected! I believe my superuser and staff user were from the same Agency when I encountered the issue described in my last comment. I'm also feeling like the earlier behavior hasn't been fully explained, but I'm unable to recreate that bug after your last changes.

jeancochrane:
@beamalsky Yeah, I'm feeling the same way about it. If the bug occurs again we can investigate more thoroughly. For now, do I have your approval to merge?

